PR #104, opened 2025-04-21
model: handle default behavior for diffuser pipelines

rebel-jongho:
# Pull Request Description

## Type of Change

- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Overview

It is complicated from an API perspective for the user to specify the batch size for every submodule in the diffuser pipeline each time. Also, the current main implementation does not work with the example code specified in model_zoo or the user docs, which used to work. 

This PR changes it to compile by referencing the default value of the Huggingface library unless a batch size is otherwise specified.

## Checklist

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

### DIFF
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -435,6 +435,7 @@ class RBLNModelConfig:
         "activate_profiler",
     ]
     submodules: List[str] = []
+    subclass_non_save_attributes = []
 
     def init_submodule_config(
         self,
@@ -463,7 +464,11 @@ class RBLNModelConfig:
         return submodule_config
 
     def __setattr__(self, key, value):
-        if key != "_attributes_map" and key not in self.non_save_attributes:
+        if (
+            key != "_attributes_map"
+            and key not in self.non_save_attributes
+            and key not in self.subclass_non_save_attributes
+        ):
             self._attributes_map[key] = value
 
         if hasattr(self, "_frozen") and self._frozen:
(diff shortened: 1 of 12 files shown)


src/optimum/rbln/configuration_utils.py
```diff
@@ -706,6 +711,29 @@ def initialize_from_kwargs(
 
         return rbln_config, kwargs
 
+    def get_default_values_for_original_cls(self, func_name: str, keys: List[str]) -> Dict[str, Any]:
+        """
+        Get default values for original class attributes from RBLNModelConfig.
+
+        Args:
+            rbln_config (RBLNModelConfig): The RBLN model configuration.
```
rebel-seinpark, line 719: This line can be removed.

src/optimum/rbln/diffusers/configurations/pipelines/configuration_stable_diffusion.py
```diff
@@ -87,11 +82,16 @@ def __init__(
             sample_size=image_size,  # image size is equal to sample size in vae
         )
 
-        if guidance_scale is not None:
-            logger.warning("Specifying `guidance_scale` is deprecated. It will be removed in a future version.")
+        # Get default guidance scale from original class to set UNet batch size
+        if guidance_scale is None:
+            guidance_scale = self.get_default_values_for_original_cls("__call__", ["guidance_scale"])["guidance_scale"]
+
+        if not self.unet.batch_size_is_specified:
```
rebel-seinpark, line 89: How do we handle the cases where the `guidance scale` has different values ​​at compile time and runtime?
For example, user gave `guidance scale=5` at compile time, `guidance scale=0` at runtime -> Dose it raise error?

src/optimum/rbln/configuration_utils.py
rebel-jongho: thanks!

src/optimum/rbln/diffusers/configurations/pipelines/configuration_stable_diffusion.py
rebel-jongho: For your question: In that case, batch size of runtime UNet will be different, So runtime raises error. 

This PR covers the case where users have not specified batch size or guidance scale.

rebel-kblee (approved):
lgtm 👍

PR #142, opened 2025-05-26
model: dynamic output class

rebel-jongho:
# Pull Request Description

## Type of Change

- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Motivation and Context

In the old code, the forward function of a model class with multiple outputs would assign multiple outputs to a single field in BaseModelOutput.

This PR fixes that.

However, there is a case where returning a tuple is still behaving differently than huggingface (flattening nested tensors), which will need to be fixed in the future.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (If needed)

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update


src/optimum/rbln/modeling.py
```diff
@@ -245,16 +244,66 @@ def forward(self, *args, return_dict: Optional[bool] = None, **kwargs):
...
+                                    return cls._output_class
+                    else:
+                        if hasattr(return_type, "__module__") and (
+                            "transformers" in return_type.__module__ or "diffusers" in return_type.__module__
+                        ):
+                            cls._output_class = return_type
+                            return cls._output_class
+            else:
+                # Fallback to BaseModelOutput
+                cls._output_class = BaseModelOutput
```
rebel-kblee, line 280: If there is no forward method, it always maps to BaseModelOutput, but I'm guessing there must be an exception case?

src/optimum/rbln/modeling.py
```diff
@@ -245,16 +244,66 @@ def forward(self, *args, return_dict: Optional[bool] = None, **kwargs):
...
-            return (output,) if not isinstance(output, (tuple, list)) else output
+            return tuple_output
         else:
-            if self.output_class is None:
-                return BaseModelOutput(last_hidden_state=output)
-
-            # Create output with the appropriate class and key
-            return self.output_class(**{self.output_key: output})
+            output_class = self.get_hf_output_class()
+            if hasattr(output_class, "loss"):
```
rebel-kblee, line 295: only for causalLM? What about other cases that might not be related to loss?

src/optimum/rbln/modeling.py
rebel-jongho: The `forward` method always exists Since transformer's PretrainedModel / diffusers model classes inherits from torch.nn.Module.

rebel-thkim: In the case of inference using the `__call__ ` method, is it possible that (diffuser) will never encounter this logic?

rebel-thkim (approved):

src/optimum/rbln/modeling.py
rebel-jongho: this is not only for causal lm. why did you think that?

src/optimum/rbln/modeling.py
rebel-jongho: In case of diffusers, they use `RBLNDiffuserMixin` (which uses `__call__`) for pipelines and RBLNModel(which uses `forward`) for its submodules. 

So, you are right, diffusers classes will never encounter this logic.

rebel-junamsong: @rebel-jongho How about writing it like this?

```
    @classmethod
    def get_hf_output_class(cls):
        if cls._output_class:
            return cls._output_class

        hf_class = cls.get_hf_class()
        if hf_class is None:
            raise ValueError(f"No HuggingFace model class found for {cls.__name__}")

        hints = get_type_hints(hf_class.forward) if hasattr(hf_class, "forward") else {}
        ret = hints.get("return")
        candidates = get_args(ret) if get_origin(ret) is Union else (ret,)

        for t in candidates:
            mod = getattr(t, "__module__", "")
            if "transformers" in mod or "diffusers" in mod:
                cls._output_class = t
                return t

        cls._output_class = BaseModelOutput
        return BaseModelOutput
```

src/optimum/rbln/modeling.py
rebel-kblee: if output_class == BaseModelOutput, BaseModelOutput has no attribute "loss". 
With this case, It does not add for tuple_output

src/optimum/rbln/modeling.py
rebel-jongho: @rebel-junamsong I apllied it, with additional changes. Thanks

PR #195, opened 2025-06-30
other: pre-check compile config

rebel-jongho:
# Pull Request Description

## Type of Change

- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Overview

If the compile function is passed compile settings that prevent it from generating runtimes on this machine, and create_runtimes=True, cause it to raise an error before compiling.

## Checklist

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

### DIFF
--- a/src/optimum/rbln/utils/runtime_utils.py
+++ b/src/optimum/rbln/utils/runtime_utils.py
@@ -41,7 +41,9 @@ def tp_and_devices_are_ok(
         if any(not isinstance(d, int) for d in device):
             return "Device must be a(n) (list of) integer(s)."
         if len(device) != tensor_parallel_size:
-            return f"The number of devices ({len(device)}) does not match tensor parallel size ({tensor_parallel_size})."
+            return (
+                f"The number of devices ({len(device)}) does not match tensor parallel size ({tensor_parallel_size})."
+            )
     else:
         return f"Invalid device: {device}"
 
--- a/src/optimum/rbln/utils/runtime_utils.py
+++ b/src/optimum/rbln/utils/runtime_utils.py
@@ -13,10 +13,46 @@
 # limitations under the License.
 
 import threading
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional, Union
 
 import rebel
 import torch
+
+
+def tp_and_devices_are_ok(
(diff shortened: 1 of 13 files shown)


src/optimum/rbln/utils/runtime_utils.py
```diff
@@ -13,12 +13,52 @@
...
+    npu: Optional[str] = None,
+) -> Optional[str]:
+    if tensor_parallel_size is None:
+        tensor_parallel_size = 1
+
+    if rebel.device_count() < tensor_parallel_size:
+        return (
+            f"Tensor parallel size {tensor_parallel_size} is greater than "
+            f"the number of available devices {rebel.device_count()}."
+        )
```
rebel-eunji, line 34: I think we need to consider two more cases.
case 1) if all of the elements of `device` are negative? ex) device = [-1, -2, -3]
case 2) if len(`device`) != `tensor_parallel_size`

tests/test_transformers.py
```diff
@@ -88,8 +88,18 @@ def test_compile_from_model(self):
...
 
-    def test_failed_to_create_runtime(self):
+    def _inner_test_save_load(self, tmpdir):
+        super()._inner_test_save_load(tmpdir)
+
         with self.assertRaises(RBLNRuntimeError):
+            _ = self.RBLN_CLASS.from_pretrained(
+                tmpdir,
+                export=False,
+                rbln_device=[0, 1, 2, 3],
```
rebel-eunji, line 98: How about `rbln_devices`(plural)?

rebel-eunji (changes requested):

src/optimum/rbln/utils/runtime_utils.py
rebel-jongho: That's a great catch! I'll fix that case.

PR #209, opened 2025-07-07
other: support runtime timeout

rebel-dkhong:
# Pull Request Description

## Type of Change

- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Overview

Add a new argument to `rbln_runtime_kwargs`.

## Motivation and Context

Large models such as Cosmos transformer need longer runtime timeout, but runtime timeout cannot be set.

## Checklist

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

### DIFF
--- a/src/optimum/rbln/diffusers/configurations/models/configuration_transformer_cosmos.py
+++ b/src/optimum/rbln/diffusers/configurations/models/configuration_transformer_cosmos.py
@@ -53,7 +53,7 @@ class RBLNCosmosTransformer3DModelConfig(RBLNModelConfig):
             ValueError: If batch_size is not a positive integer.
         """
         if kwargs["timeout"] is None:
-            kwargs["timeout"] = 120
+            kwargs["timeout"] = 80
 
         super().__init__(**kwargs)
         self.batch_size = batch_size or 1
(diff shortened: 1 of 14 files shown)


rebel-jongho (changes requested):
Well, the concept of `timeout` is not dependent on the model class, but rather on the Pretrained model itself. For example, any large (compute-heavy) model could theoretically exceed the predefined `timeout`.

That said, I think `timeout` should be a runtime-configurable value, and therefore should be handled at the same level as `rbln_device` or `rbln_create_runtimes`.

Of course, as you considered, some models (like **cosmos-vae**) may require a high `timeout` value in the default model that the user is using, so I agree that the default-timeout value may need to be set per class.

Putting all this together, I think it would be better to specify the default value of `timeout` in the rbln_config class instead of the rbln_model class.

like `timeout=rbln_config.runtime_timeout`

src/optimum/rbln/diffusers/configurations/models/configuration_transformer_cosmos.py
```diff
@@ -52,6 +52,9 @@ def __init__(
         Raises:
             ValueError: If batch_size is not a positive integer.
         """
+        if kwargs["timeout"] is None:
+            kwargs["timeout"] = 80
```
rebel-jongho, line 56:
```suggestion
        if kwargs.get("timeout") is None:
```
or use explicit arg of `__init__` func

PR #550, opened 2026-04-20
doc(t5): fix Optimum RBLN Overview link for MkDocs

rebel-junamsong:
## Summary
T5 model docstrings used `../../../optimum_rbln.md`, which does not exist in the RBLN SDK documentation build. Linkchecker on `rbln_sdk_doc` then reported 404s from the generated T5 API pages.

## Change
Point the **Optimum RBLN Overview** link to `../../../index.md`, which resolves to `software/optimum/index.md` in the built site (same section as the main Optimum RBLN overview).

## Note
Rebased onto current `dev` (replaces the previous branch tip that was based on an older commit).


rebel-kblee (approved):
👍

PR #569, opened 2026-05-22
model: lower minimum flash attention seq len and partition length

rebel-thkim:
## Summary
- Lower `MIN_FLASH_ATTN_MAX_SEQ_LEN` from 8192 to 2048
- Lower `MIN_FLASH_ATTN_PARTITION_LENGTH` from 4096 to 1024

Allows flash attention to be configured with smaller sequence lengths and partition lengths.

## Test plan
- [x] Verify flash attention compiles and runs with the new minimum values
- [x] Confirm existing models using default values are unaffected


rebel-kblee (approved):
check

PR #576, opened 2026-05-29
other: Revert "model: lower minimum flash attention seq len and partition length"

rebel-thkim:
Reverts RBLN-SW/optimum-rbln#569


rebel-jswoo (approved):

rebel-shshin (approved):
👍

PR #597, opened 2026-06-18
refactor: rename tensor_parallel_size to num_devices

rebel-jongho:
# Pull Request Description

Since the compiler's `tensor_parallel_size` has been renamed to `num_devices`, we are also renaming the `tensor_parallel_size` used here. However, for backward compatibility, we are marking it as deprecated. 

## Type of Change
- [x] Code refactoring

## Changes Overview
rebel-compiler renamed its public compile argument `tensor_parallel_size` to the parallelization-neutral `num_devices`, keeping the old name as a deprecated alias (SR-15, rebel_compiler#11330). This PR ports the same rename into optimum-rbln.

- Renamed the public knob `tensor_parallel_size` to `num_devices` on `RBLNModelConfig` and `RBLNCompileConfig`. The old name is kept as a deprecated alias via `deprecate_kwarg`, so both `tensor_parallel_size` and `rbln_tensor_parallel_size` keep working (they map to `num_devices` with a deprecation notice).
- At the compiler boundary (`rebel.compile_from_torch` in `modeling_base.py`), feature-detect whether `rebel.compile` accepts `num_devices` via `inspect.signature`. If it does, pass `num_devices`; otherwise fall back to `tensor_parallel_size`. This avoids version-string comparison, so it works against both new and pre-rename compilers and stays correct after the alias is eventually removed.
- Models compiled with an older version still load: when reconstructing `RBLNCompileConfig` from a serialized `rbln_config.json` (whose compile cfgs carry `tensor_parallel_size`), unknown keys are dropped and `num_devices` is repopulated from the parent config.
- Updated CLI help, docstrings, example scripts, and tests to the new name, and added a test asserting the deprecated alias maps to `num_devices`.

Since `num_devices` is not specific to tensor parallelism, user-facing wording was changed from "tensor parallel size" to "number of devices". The internal `_tp_support` flag was left as-is, because it genuinely describes a model's tensor-parallelism support.

## Motivation and Context
Keeps optimum-rbln's public API aligned with the rebel-compiler rename (SR-15). The deprecated alias ensures existing user code and previously compiled artifacts do not break.

## Related Issues
Related to SR-15 (rebel_compiler#11330)

----
# Conventional commit
```
refactor: rename tensor_parallel_size to num_devices
```

### DIFF
--- a/examples/image-to-text/run_idefics3.py
+++ b/examples/image-to-text/run_idefics3.py
@@ -14,7 +14,7 @@ def main(
     from_transformers: bool = False,
     prompt: typing.Optional[str] = None,
     max_seq_len: typing.Optional[int] = None,
-    tensor_parallel_size: typing.Optional[int] = 4,
+    num_devices: typing.Optional[int] = 4,
 ):
     processor = AutoProcessor.from_pretrained(model_id)
 
@@ -27,7 +27,7 @@ def main(
                     "attn_impl": "flash_attn",
                     "max_seq_len": max_seq_len,
                     "use_inputs_embeds": True,
-                    "tensor_parallel_size": tensor_parallel_size,
+                    "num_devices": num_devices,
                     "batch_size": batch_size,
                 }
             },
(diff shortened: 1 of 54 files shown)


rebel-eunji (approved):

rebel-thkim (approved):
👍

PR #599, opened 2026-06-19
model: support max_seq_len bucketing for transformer encoders

rebel-thkim:
## What

Adds **sequence-length bucketing** to `RBLNTransformerEncoder` (and therefore all feature-extraction encoders built on it — `RBLNBertModel`, `RBLNXLMRobertaModel`, `RBLNT5EncoderModel`, `RBLNBartModel`, `RBLNPegasusModel`, …).

`max_seq_len` now accepts a list:

```python
RBLNBertModel.from_pretrained(model_id, export=True, rbln_max_seq_len=[512, 1024, 2048, 4096])
```

All buckets are compiled into a **single** model (multiple `input_info` sets → one `compile_from_torch` call), so the rebel compiler shares one copy of the encoder weights across buckets (`device_func_count == 1`). At inference the `BucketingSyncRuntime` dispatches by input shape.

## How

- **`configuration_generic.py`**: `max_seq_len: Optional[Union[int, List[int]]]`. A list is mutually exclusive with `model_input_shapes` (which fully specifies a single graph) — combining them raises `ValueError`.
- **`modeling_generic.py`**:
  - `update_rbln_config_for_transformers_encoder` builds one `input_info` set per bucket (sorted/deduped); a single bucket keeps the original flat form (no behavior change).
  - `forward` routes an arbitrary-length input to the smallest bucket that fits, pads up to it, runs, then slices the sequence dimension of the outputs back. Routing only activates when multiple buckets are compiled, so single-length models are byte-for-byte unchanged.

## Tests

`tests/test_transformers.py::TestEncoderMaxSeqLenBucketing` — compiles tiny BERT (e5 path) and XLM-RoBERTa (bge path) with `[32, 64, 128]` buckets, asserts one executor per bucket, and checks routed outputs match the HF reference at lengths spanning sub-/exact-/between-bucket cases; plus a unit test for the `max_seq_len` + `model_input_shapes` conflict.

### DIFF
--- a/src/optimum/rbln/transformers/configuration_generic.py
+++ b/src/optimum/rbln/transformers/configuration_generic.py
@@ -22,7 +22,7 @@ class RBLNTransformerEncoderConfig(RBLNModelConfig):
 
     def __init__(
         self,
-        max_seq_len: Optional[int] = None,
+        max_seq_len: Optional[Union[int, List[int]]] = None,
         batch_size: Optional[int] = None,
         model_input_names: Optional[List[str]] = None,
         model_input_shapes: Optional[List[Tuple[int, int]]] = None,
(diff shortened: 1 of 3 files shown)


src/optimum/rbln/transformers/modeling_generic.py
```diff
@@ -180,6 +196,39 @@ def update_rbln_config_for_transformers_encoder(
...
+        compile_cfg = self.rbln_config.compile_cfgs[0]
+        if not compile_cfg.is_multiple_input_info:
+            # No sequence-length bucketing: use the default single-shape path.
+            return super().forward(*args, return_dict=return_dict, **kwargs)
+
+        # Bucketing: pad inputs to the smallest compiled `max_seq_len` that fits (the runtime then
+        # dispatches to the matching executor), and slice the sequence dimension of outputs back.
+        # Multiple input_info sets only ever come from list-valued `max_seq_len` (the config forbids
+        # combining it with `model_input_shapes`), so index [1] is always the sequence dimension.
+        buckets = sorted(info[0][1][1] for info in compile_cfg.input_info)
```
rebel-thkim: fragile

src/optimum/rbln/transformers/modeling_generic.py
```diff
@@ -180,6 +196,39 @@ def update_rbln_config_for_transformers_encoder(
...
+        def pad(t):
+            return torch.nn.functional.pad(t, (0, target - seq_len)) if isinstance(t, torch.Tensor) else t
+
+        def unpad(t):
+            return t[:, :seq_len] if isinstance(t, torch.Tensor) and t.dim() >= 2 and t.shape[1] == target else t
+
+        output = self.model[0](*(pad(a) for a in args), **{k: pad(v) for k, v in kwargs.items()})
+        output = type(output)(map(unpad, output)) if isinstance(output, (tuple, list)) else unpad(output)
+
+        if self.hf_library_name == "transformers":
```
rebel-thkim: why u need it?

src/optimum/rbln/transformers/configuration_generic.py
```diff
@@ -22,24 +22,40 @@ class RBLNTransformerEncoderConfig(RBLNModelConfig):
...
+                scalar `max_seq_len` (or `None`) when using it.
             kwargs: Additional arguments passed to the parent RBLNModelConfig.
 
         Raises:
-            ValueError: If batch_size is not a positive integer.
+            ValueError: If batch_size is not a positive integer, or if a list-valued `max_seq_len`
+                (bucketing) is combined with `model_input_shapes`.
         """
         super().__init__(**kwargs)
+        if model_input_shapes is not None and isinstance(max_seq_len, (list, tuple)):
```
rebel-kblee: deprecate

src/optimum/rbln/transformers/configuration_generic.py
```diff
@@ -22,24 +22,40 @@ class RBLNTransformerEncoderConfig(RBLNModelConfig):
...
+            max_seq_len (Optional[Union[int, List[int]]]): Maximum sequence length supported by the model.
+                A single integer compiles the model for one sequence length. A list of integers enables
+                sequence-length bucketing: the model is compiled for each length on the default
+                ``[batch_size, max_seq_len]`` input layout and, at inference time, the runtime dispatches
+                to the bucket matching the input shape. Bucketing is mutually exclusive with
+                `model_input_shapes` (which fully specifies a single set of shapes).
             batch_size (Optional[int]): The batch size for inference. Defaults to 1.
             model_input_names (Optional[List[str]]): Names of the input tensors for the model.
                 Defaults to class-specific rbln_model_input_names if not provided.
+            model_input_shapes (Optional[List[Tuple[int, int]]]): Explicit shape for each input tensor,
```
rebel-thkim: deprecated

rebel-kblee (approved):
lgtm 👍

PR #600, opened 2026-06-22
performance: normalize weights to contiguous before compile to cut host peak memory

rebel-jongho:
## Type of Change
- [x] Performance improvement
- [ ] Breaking change
- [ ] Documentation update
- [ ] Code refactoring

## Changes Overview
Add a `normalize_contiguous_` helper that materializes every parameter and buffer as a contiguous tensor in place, and call it right before `rebel.compile_from_torch`:
- from `RBLNBaseModel.compile`, the chokepoint every standard compile path funnels through (`RBLNModel`, seq2seq/T5, decoder-only `_compile_model`);
- from the two direct `compile_from_torch` calls in the Cosmos guardrail pipeline.

Already-contiguous weights (every safetensors checkpoint) are left untouched, so this is a no-op for the common case.

## Motivation and Context
`transformers` v5's `core_model_loading` loader can preserve a linear weight as a **non-contiguous transposed view** (stride `(1, N)`) when a `.bin` checkpoint stores it that way. `transformers` v4 normalized the same `.bin` to contiguous on load.

The compiler calls `.contiguous()` on every weight at ingest — a zero-copy no-op for a contiguous weight, but a full second buffer for each non-contiguous one. Those copies coexist with the originals still held by the `ExportedProgram`, so host peak memory at compile time roughly **doubles over the affected weights**.

Only models whose `.bin` checkpoints carry transposed (non-square) linear weights are affected (e.g. t5-11b's encoder/decoder projections); safetensors checkpoints are always materialized contiguous, and square-weight models are unaffected.

Replacing each tensor in place frees its original immediately, so the transient cost is a single weight instead of the whole set, and the compiler's later `.contiguous()` becomes a no-op.

**Validation.** Controlled reproduction with real `.contiguous()` semantics (12 non-contiguous transposed weights, 0.5GB each, originals held like an `ExportedProgram` while a constant pool is built):

| | after build | after make-contiguous | peak during ingest |
|---|---|---|---|
| **before** (no fix) | 6.22 GB | — | **12.22 GB** |
| **after** (this PR) | 6.22 GB | 6.72 GB | **6.72 GB** |

The doubling over the non-contiguous mass is eliminated; the only added cost is a single-weight transient. Confirmed that `torch.save`/`load` of a `.bin` preserves the non-contiguous stride while safetensors does not.

Helper unit-tested for: contiguity, value preservation, **weight-tying preservation** (tied params stay the same object), idempotency, and **data_ptr stability** across repeated calls (so weight-sharing across multi-phase compiles is unaffected).

## Related Issues
N/A

----
# Conventional commit
```
performance: normalize weights to contiguous before compile to cut host peak memory
```

### DIFF
--- a/src/optimum/rbln/diffusers/pipelines/cosmos/cosmos_guardrail.py
+++ b/src/optimum/rbln/diffusers/pipelines/cosmos/cosmos_guardrail.py
@@ -25,6 +25,7 @@ from huggingface_hub import snapshot_download
 from transformers import AutoTokenizer, SiglipProcessor
 
 from .... import RBLNAutoModelForCausalLM, RBLNSiglipVisionModel
+from ....modeling_base import normalize_contiguous_
 from ....utils.runtime_utils import RBLNPytorchRuntime, UnavailableRuntime
 from .configuration_cosmos_guardrail import RBLNCosmosSafetyCheckerConfig
 
@@ -167,6 +168,7 @@ class RBLNRetinaFaceFilter(RetinaFaceFilter):
                 super().__init__(checkpoint_id)
             net = self.net
             del self.net
+            normalize_contiguous_(net)
             self.compiled_model = rebel.compile_from_torch(
                 net,
                 input_info=[
@@ -245,6 +247,7 @@ class RBLNVideoSafetyModel(VideoSafetyModel):
             safety_filter_local_path = os.path.join(checkpoint_dir, "safety_filter.pt")
             checkpoint = torch.load(safety_filter_local_path, weights_only=True)
             network.load_state_dict({k.replace("network.", ""): v for k, v in checkpoint["model"].items()})
+            normalize_contiguous_(network)
 
             self.compiled_model = rebel.compile_from_torch(
                 network,
--- a/src/optimum/rbln/modeling_base.py
+++ b/src/optimum/rbln/modeling_base.py
@@ -38,6 +38,20 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 
 
+def normalize_contiguous_(model: torch.nn.Module) -> torch.nn.Module:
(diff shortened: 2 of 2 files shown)


rebel-thkim:
It looks like there's a part in `cosmos_guardrail` that directly calls `compile_from_torch`.

rebel-jongho:
Good catch — those two direct `compile_from_torch` calls in `cosmos_guardrail.py` are already covered in this PR:

- `RBLNRetinaFaceFilter`: `normalize_contiguous_(net)` at line 171, right before `compile_from_torch(net, ...)` at line 172.
- `RBLNVideoSafetyModel`: `normalize_contiguous_(network)` at line 250, right before `compile_from_torch(network, ...)` at line 252.
- Import added at line 28 (`from ....modeling_base import normalize_contiguous_`).

The only other caller is `RBLNBaseModel.compile`, which is also covered. A repo-wide grep for `compile_from_torch` confirms these are the only three call sites.

rebel-kblee:
- If this is an issue with loading the Transformers model, does it always occur when using `rebel.compile_from_torch` or `torch.compile` in contexts other than `optimum`?

- Secondly, is there any overhead (?) associated with this? Is there no need to apply it in cases where the data is already contiguous?

rebel-jongho:
> * If this is an issue with loading the Transformers model, does it always occur when using `rebel.compile_from_torch` or `torch.compile` in contexts other than `optimum`?

This phenomenon always occurs when the model’s weights are not contiguous. It is not a problem specific to `transformers` or `optimum-rbln`.

> * Secondly, is there any overhead (?) associated with this? Is there no need to apply it in cases where the data is already contiguous?

contiguous() will not be effective when it is already contiguous. 

if it's not contiguous, then there will be memory allocation and copy.

rebel-kblee (approved):
lgtm 👍

PR #612, opened 2026-06-26
other(auto): use RBLNAutoConfig.from_pretrained instead of removed .load

rebel-junamsong:
## Problem

Loading any compiled model through an `RBLNAutoModelFor*` class (`export=False`) crashes on 0.11.0:

```
File ".../optimum/rbln/transformers/models/auto/auto_factory.py", line 177, in get_rbln_model_cls_name
    rbln_config = RBLNAutoConfig.load(model_path_subfolder)
ValueError: `RBLNAutoConfig.load` is deprecated and removed starting from version 0.11.0. Use `from_pretrained` instead.
```

`RBLNAutoConfig.load` is decorated with `@deprecate_method(version="0.11.0", new_method="from_pretrained")` and now hard-raises — but `AutoFactory.get_rbln_model_cls_name` still calls it internally. So the auto-class load path calls a method the library itself removed.

- **Compile (`export=True`)** is unaffected — the class is resolved from the HF `AutoConfig`, not from `RBLNAutoConfig.load`.
- **Inference / load of a compiled model (`export=False`)** hits this path and fails.
- Affects every `RBLNAutoModelFor*` (e.g. `SpeechSeq2Seq`/Whisper, `TextEncoding`/BGE, `SequenceClassification`/reranker).

Surfaced by the rbln-deploy nightly suite (ko-reranker): compile RC 0, inference RC 1.

## Fix

Call the non-deprecated `from_pretrained`, which is a drop-in (identical signature and return type — `load` is just a deprecated alias):

```diff
- rbln_config = RBLNAutoConfig.load(model_path_subfolder)
+ rbln_config = RBLNAutoConfig.from_pretrained(model_path_subfolder)
```

Also updates a stale `RBLNAutoConfig.load(...)` usage example in `configuration_utils.py`. The `load()` method's own "Deprecated usage" example is intentionally left unchanged.


rebel-seinpark (approved):
LGTM

PR #621, opened 2026-06-29
refactor(ops): drop duplicate custom op registration, defer to rebel-compiler

rebel-jongho:
# Pull Request Description

## Type of Change
- [x] Code refactoring

## Changes Overview

optimum-rbln kept its own copy of the `rbln_custom_ops::*` custom op definitions under `src/optimum/rbln/ops/`, but rebel-compiler already registers the same ops (a strict superset) at import time. Both ran on `import optimum.rbln`, so the effective op schema depended on load order. This PR makes rebel-compiler the single source of truth and removes optimum's duplicate registrations.

- Delete `src/optimum/rbln/ops/` (`attn`, `flash_attn`, `kv_cache_update`, `linear`, `moe`). rebel-compiler registers all of these ops and is eagerly imported via `utils/runtime_utils.py`, so registration stays guaranteed on `import optimum.rbln`.
- Move the non-registration MoE routing helpers (`compute_masked_routing_weight_softmax_first` / `_topk_first`) to `transformers/utils/moe.py`, and repoint the 6 MoE architectures (mixtral, qwen2_moe, qwen3_moe, qwen3_vl_moe, gpt_oss, gemma4) to the new location.
- Drop the side-effect op imports from 6 model `__init__` files (bart, pegasus, whisper, t5, time_series_transformer, decoderonly). Every call site already dispatches through `torch.ops.rbln_custom_ops.*`, so no call site changes are needed.

## Motivation and Context

Two independent registrations of the same custom op name is fragile: it relies on import order and silently overrides, and the two schemas had already drifted in the past. Keeping registration in rebel-compiler alone removes that ambiguity and is the first step toward a single source of truth for custom ops (SR-101).

The op-name coverage is verified: rebel-compiler registers all 22 ops that optimum previously registered (plus `paged_sliding_window_attn_*`), the MoE signatures are canonical and compatible (`custom_moe_ff` byte-identical; `custom_moe_glu` / `custom_moe_glu_mxfp4` rebel-side are supersets with trailing optional args), and `import optimum.rbln` still registers all 22 ops with the affected model modules importing cleanly.

## Related Issues

SR-101


rebel-kblee (approved):
lgtm 👍

PR #625, opened 2026-07-07
model: add RBLNModernBertForMaskedLM

ho4040:
## Type of Change
- [x] New Model Support

## Changes Overview
Adds `RBLNModernBertForMaskedLM` (+ `RBLNModernBertForMaskedLMConfig`) for the
ModernBERT encoder, following the existing `RBLNRobertaForMaskedLM` pattern.

ModernBERT needs two model-specific adjustments over the generic encoder path:

- **Force SDPA at load** (`get_pytorch_model`). ModernBERT selects its attention
  backend from `config._attn_implementation` and prefers HuggingFace's
  `flash_attention_2` (the CUDA-only `flash-attn` kernels) when that package is
  present; rebel-compiler cannot compile that path for RBLN. We pin
  `attn_implementation="sdpa"` so compilation is deterministic regardless of the
  host environment and the compiler lowers `scaled_dot_product_attention` to its
  own kernel. *(This is unrelated to RBLN's own `flash_attn` `attn_impl`, which
  is a decoder KV-cache feature and does not apply to encoders.)*
- **A dedicated compile wrapper** (`ModernBertModelWrapper`) forwards the 2D
  padding mask unchanged instead of pre-expanding it with
  `_prepare_4d_attention_mask`. ModernBERT alternates full and local
  (sliding-window) attention and builds both 4D masks internally via
  `create_bidirectional_mask` / `create_bidirectional_sliding_window_mask`;
  pre-expanding would feed the global mask to the sliding layers and drop the
  local windowing.

The class is registered in the three `__init__` export lists;
`RBLNAutoModelForMaskedLM` resolves it through the existing naming convention.
A test mirroring `TestBertForMaskedLM` is added
(`hf-internal-testing/tiny-random-ModernBertForMaskedLM`).

## Motivation and Context
ModernBERT (including the multilingual **mmBERT**) currently has no official
optimum-rbln support. We built and validated this adapter end-to-end on real
ATOM+ hardware:

- **Env:** ATOM+ (RBLN-CA22), rebel-compiler 0.11.0, optimum-rbln 0.11.0.post1,
  transformers 5.8.1, torch 2.11.0
- **Compile:** `jhu-clsp/mmBERT-base`, `export=True`, `max_seq_len=128`,
  `batch_size=1`, fp32 → `compiled_model.rbln` produced
- **Accuracy vs CPU eager (SDPA):** top-1 agreement = **1.00** over all real
  token positions; masked-token prediction matches (`"Paris"`); masked-slot
  logit cosine similarity 0.957

We also published a technical write-up of the port, showcasing running an
unsupported encoder on ATOM+:
- English: https://warpspaceinc.github.io/blog/posts/modernbert-on-rebellions-npu/
- 한국어: https://warpspaceinc.github.io/blog/ko/posts/modernbert-on-rebellions-npu/

We'd love ModernBERT to become officially supported so we can point readers at
first-party support. Happy to adjust to your conventions, extend sequence-length
/ bucketing coverage, or hand off for an accuracy review.


rebel-kblee (approved):
LGTM 👍 Thanks for contribute!

PR #631, opened 2026-07-14
model: Qwen3.5

rebel-seinpark:
# Pull Request Description

> 

## Type of Change

- [x] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update


src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
```diff
@@ -0,0 +1,214 @@
...
+            visual (Optional[RBLNModelConfig]): Configuration for the vision encoder submodule.
+            _load_visual_runtime (bool): Whether to create the visual encoder runtime. Set False on
+                decoder-only nodes in a disaggregated setup (then pre-computed image_embeds must be fed to
+                forward()). Defaults to True.
+            kwargs: Additional arguments passed to `RBLNDecoderOnlyModelForCausalLMConfig`.
+
+        Raises:
+            ValueError: If `use_inputs_embeds` is False.
+        """
+        super().__init__(use_inputs_embeds=use_inputs_embeds, **kwargs)
```
rebel-thkim, line 207: 207~214 lines are redundant with super().__init__, how about fully inherit __init__ from ~Model

src/optimum/rbln/transformers/models/qwen3_5/__init__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-jongho: 2026

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
```diff
@@ -0,0 +1,214 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-jongho: 2026

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
```diff
@@ -0,0 +1,214 @@
...
+        self.gdn_chunk_size = gdn_chunk_size
+
+
+class RBLNQwen3_5VisionModelConfig(RBLNModelConfig):
+    """Vision encoder config for Qwen3.5 (Qwen3-VL-style vision, no deepstack): per-image `max_seq_len`.
+
+    Independent of the Qwen3-VL config (inherits `RBLNModelConfig` directly).
+    """
+
+    @deprecate_kwarg(old_name="max_seq_lens", new_name="max_seq_len", version="0.11.0")
```
rebel-jongho: already exceeded version 0.11

rebel-jongho: you may find this PR also related https://github.com/RBLN-SW/optimum-rbln/pull/644

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
```diff
@@ -0,0 +1,214 @@
...
+
+    Qwen3.5 is natively vision-language, so this is the multimodal model config. Independent of the
+    Qwen3-VL config (inherits `RBLNDecoderOnlyModelConfig` directly), carrying its own `visual`
+    submodule handling plus the Qwen3.5-specific `gdn_chunk_size`. Which layers are linear is read
+    from the HF `config.text_config.layer_types`. The vision encoder output is injected into
+    `inputs_embeds` (`use_inputs_embeds=True`).
+    """
+
+    submodules = ["visual"]
+    subclass_non_save_attributes = ["_load_visual_runtime", "memory_budget"]
```
rebel-jongho, line 123: 👍 Great

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+        )
+
+        model.save_pretrained("compiled-qwen3.5-0.8b")
+        ```
+    """
+
+    auto_model_class = AutoModelForCausalLM
+
+    def forward(self, *args, **kwargs):
+        kwargs["return_dict"] = True
```
rebel-thkim: Is the fact that `return_dict=True` is hardcoded consistent with how Transformers work?

src/optimum/rbln/transformers/__init__.py
```diff
@@ -182,6 +182,14 @@
...
         "RBLNQwen3ModelConfig",
+        "RBLNQwen3_5Model",
+        "RBLNQwen3_5ModelConfig",
+        "RBLNQwen3_5ForCausalLM",
+        "RBLNQwen3_5ForCausalLMConfig",
+        "RBLNQwen3_5ForConditionalGeneration",
+        "RBLNQwen3_5ForConditionalGenerationConfig",
+        "RBLNQwen3_5VisionModel",
+        "RBLNQwen3_5VisionModelConfig",
         "RBLNQwen3VLForConditionalGeneration",
```
rebel-jongho, line 193: keep list sorted

src/optimum/rbln/__init__.py
```diff
@@ -187,6 +187,14 @@
...
         "RBLNQwen3ModelConfig",
+        "RBLNQwen3_5Model",
+        "RBLNQwen3_5ModelConfig",
+        "RBLNQwen3_5ForCausalLM",
+        "RBLNQwen3_5ForCausalLMConfig",
+        "RBLNQwen3_5ForConditionalGeneration",
+        "RBLNQwen3_5ForConditionalGenerationConfig",
+        "RBLNQwen3_5VisionModel",
+        "RBLNQwen3_5VisionModelConfig",
         "RBLNQwen3VLVisionModel",
```
rebel-jongho, line 198: please keep the list sorted

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+        conv_state_len = text_config.linear_conv_kernel_dim - 1
+
+        kvcache_dtype = rbln_config.dtype
+        if rbln_config.quantization and rbln_config.quantization.kv_caches == "fp8":
+            kvcache_dtype = "float8_e4m3fn"
+
+        # The linear conv/recurrent caches are ONE shared static tensor across prefill and every decode
+        # bucket, so they are sized to the max (decode) batch — NOT the per-config `batch_size` (=1 for
+        # prefill). Prefill processes one item and read/writes its `batch_position` slot; decode runs the
+        # full batch. (Like the paged KV cache, which is block-indexed and likewise batch-config-agnostic.)
```
rebel-thkim: Some of the comments seem unnecessarily long. It would be great if you could go through and organize the comments in this PR.

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
rebel-seinpark: The two configs sit on different bases on purpose:

- `RBLNQwen3_5ModelConfig` -> `RBLNDecoderOnlyModelConfig` (bare model)
- `RBLNQwen3_5ForConditionalGenerationConfig` -> `RBLNDecoderOnlyModelForCausalLMConfig` (generation model)

`RBLNDecoderOnlyModelForCausalLMConfig` adds the generation-critical defaults:

```
_default_phases = ["prefill", "decode"]
_default_logits_to_keep = 1
```

Re-parenting the generation config to `RBLNQwen3_5ModelConfig` drops `RBLNDecoderOnlyModelForCausalLMConfig` from the chain, so it loses the "`decode`" phase (-> no `decode` runtime, generation breaks) and `logits_to_keep` falls back to 0.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,230 @@
...
+            if block_tables is not None:
+                named["block_tables"] = block_tables
+            if position_embed_chunk is not None:
+                named["position_emb"] = position_embed_chunk
+            if self.rbln_config.logits_to_keep > 0:
+                named["query_position"] = query_position
+            if self.rbln_config.use_attention_mask:
+                named["attention_mask"] = chunked_attention_mask
+            if self.rbln_config.use_lora:
+                named["lora_int_ids"] = lora_int_ids
```
rebel-jongho, line 149: **LoRA int-ids are never resolved.**

The base `RBLNRuntimeModel.prefill_forward` resolves `lora_int_ids` from `self.lora_int_ids` when the caller does not pass them, and raises a clear error when neither is set (`decoderonly_runtime_utils.py:520-530`). This override drops that block, so a caller that used `set_lora_int_ids()` reaches here with `lora_int_ids=None` and feeds `None` into the graph.

The path is otherwise live — `get_input_info` adds the `lora_int_ids` input whenever `rbln_config.use_lora` — so this is reachable rather than dead. Please carry the base's resolution over (or delegate to it).

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,230 @@
...
+            position_embed,
+            padded_cache_lengths,
+            query_length,
+            token_type_ids,
+        ) = self._prepare_prefill_inputs(
+            inputs, cache_position, attention_mask, position_ids, position_embed, token_type_ids=token_type_ids
+        )
+
+        chunk = self.rbln_config.prefill_chunk_size
+        prefix_cached_len = cache_position[0][0].item()
```
rebel-jongho, line 110: **Prefix caching is dropped.**

The base handles a non-zero `prefix_cached_len` in two ways (`decoderonly_runtime_utils.py:549-558`): it rejects a value that is not a multiple of `prefill_chunk_size` with `NotImplementedError`, and it pre-reveals the cached prefix in `chunked_attention_mask`.

Here `prefix_cached_len` is read and used in the per-chunk offsets, but neither the guard nor the pre-reveal is carried over. With externally supplied prefix caching the cached prefix therefore stays masked out and the window computes against a wrong mask silently. Please port both, or reject `prefix_cached_len > 0` explicitly until it is supported.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,230 @@
...
+
+            query_position = (
+                torch.tensor(
+                    (query_length - 1) % chunk if step + chunk >= query_length else chunk - 1, dtype=torch.int16
+                )
+                if self.rbln_config.logits_to_keep > 0
+                else None
+            )
+
+            named = {"inputs_embeds" if self.rbln_config.use_inputs_embeds else "input_ids": input_chunk}
```
rebel-jongho, line 138: **`position_ids` is never passed to the graph.**

`named` is built without a `position_ids` entry, but `get_input_info` adds a `position_ids` input whenever `rbln_config.use_position_ids` (`modeling_qwen3_5.py:196-197`), and the attention-mask branch a few lines below in this same method keys off `use_position_ids` — so the option is clearly expected to be settable.

Because `_run` maps by name (`args = [named_inputs[order[k]] ...]`), enabling `use_position_ids` fails with `KeyError: 'position_ids'` rather than anything diagnosable. The base chunks it as `position_ids_chunk` and passes it positionally (`decoderonly_runtime_utils.py:566, 613`). Please pass it here too, or reject `use_position_ids` in `_update_rbln_config`.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+                attention_mask=attention_mask[b_idx] if attention_mask is not None else None,
+                cache_position=cache_position,
+                batch_idx=b_idx,
+                position_embed=position_embed[:, b_idx : b_idx + 1],
+                block_tables=self.block_tables,
+            )
+            logits.append(output.logits)
+            if self.rbln_config.output_hidden_states:
+                for l_idx in range(self.config.text_config.num_hidden_layers + 1):
+                    all_hidden_states[l_idx][b_idx].copy_(output.hidden_states[l_idx][0])
```
rebel-jongho: **`output_hidden_states` cannot work as wired.**

This line reads `output.hidden_states[l_idx][0]`, but `RBLNQwen3_5RuntimeModel` returns `hidden_states=None` unconditionally from both `prefill_forward` and `decode_forward` (`qwen3_5_runtime_utils.py:180, 230`), and `get_input_info` never adds the hidden-state outputs.

`_validate_output_hidden_states` only checks that the call argument agrees with the compiled config, so compiling with `output_hidden_states=True` gets past validation and then fails here with a `TypeError`. Either produce the hidden states in the runtime, or reject the option at config time so the failure is explicit.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+                    rbln_config.dtype,
+                )
+            )
+            # per-token validity (1=real, 0=right-padding); host-built, GatedDeltaNet uses it to drop padding.
+            input_info.append(("valid_mask", [batch_size, query_length, 1], rbln_config.dtype))
+            # prefill only: which max-batch slot this per-item (batch=1) call reads/writes in the linear caches.
+            if is_prefill:
+                input_info.append(("batch_idx", [], "int16"))
+
+        if len(rbln_config.kvcache_metas) == 0:
```
rebel-jongho: **Diverges from the base `kvcache_metas` contract.**

The base `get_input_info` branches up front: when `rbln_config.kvcache_metas` is already populated it replays `meta.compile_shape` / `meta.dtype` for the input specs, and only builds fresh metas otherwise (`modeling_decoderonly.py:398-427`).

This version always recomputes the per-layer specs and then stores the metas only if the list happens to be empty. The two agree today because everything is re-derived from the same `rbln_config`, but the stored metas stop being the single source of truth for the input shapes, so any later step that adjusts a meta (the way block-count resolution adjusts the paged KV entries) would be ignored on the second call. Please follow the base's branch shape.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,230 @@
...
+            named["position_emb"] = position_embed
+        if self.rbln_config.use_attention_mask:
+            named["attention_mask"] = attention_mask
+        if self.rbln_config.use_lora:
+            named["lora_int_ids"] = lora_int_ids
+
+        # Decode always continues from the real carried state -> ones (no-op). The masks are gated on the
+        # prefill phase in the GatedDeltaNet, so they are pruned from the decode graph and these entries
+        # are simply ignored by the name-based input mapping; passed for safety if they survive.
+        named["conv_state_mask"] = torch.ones(self.conv_state_shape, dtype=self.state_dtype)
```
rebel-jongho: **Three tensors allocated and discarded on every decode step.**

`conv_state_mask`, `recurrent_state_mask` and `valid_mask` are built fresh on each `decode_forward` call, i.e. once per generated token. The comment directly above states these are pruned from the decode graph, and `_run` only picks the names present in `self.runtime._index_to_input_name`, so they are indeed dropped — the allocation is pure waste.

`conv_state_mask` alone is `(B, K-1, 2*Hk*Dk + Hv*Dv)`, which is tens of thousands of elements per token on a large config. Please either drop these three entries from the decode path, or build them once in `__init__` and reuse them (they only ever hold zeros or ones). The same applies to the per-chunk `fill(...)` and `valid_mask` allocations in `prefill_forward`, which likewise take only two distinct values.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+    """
+
+    auto_model_class = AutoModelForCausalLM
+
+    def forward(self, *args, **kwargs):
+        kwargs["return_dict"] = True
+        return super().forward(*args, **kwargs)
+
+
+class RBLNQwen3_5VisionModel(RBLNModel):
```
rebel-jongho, line 313: **Question: is the independence from `RBLNQwen3VLVisionModel` deliberate?**

Comparing this class method-by-method against `RBLNQwen3VLVisionModel`, `rot_pos_emb` (37 lines), `fast_pos_embed_interpolate` (60 lines) and `_pad_hidden_states` (25 lines) are byte-identical, and `__post_init__` (88%), `save_torch_artifacts` (82%) and `_update_rbln_config` (78%) are near-identical. `forward` (77%) is the only method that meaningfully differs, because Qwen3.5 drops deepstack.

Dropping deepstack is a subtraction, which subclassing handles naturally: inherit `RBLNQwen3VLVisionModel` and override `forward`. As written, a fix to any of the shared helpers in qwen3_vl will not reach here.

If you deliberately want the two towers decoupled so Qwen3.5 is not exposed to qwen3_vl churn, that is a reasonable call — say so and leave it as is.

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
```diff
@@ -0,0 +1,214 @@
...
+        gdn_chunk_size: int | None = None,
+        **kwargs: Any,
+    ):
+        super().__init__(**kwargs)
+        # GDN prefill sub-chunk size (must divide prefill_chunk_size). None -> = prefill_chunk_size
+        # (n_chunks == 1, no sub-chunking). See rbln_chunk_gated_delta_rule.
+        self.gdn_chunk_size = gdn_chunk_size
+
+
+class RBLNQwen3_5VisionModelConfig(RBLNModelConfig):
```
rebel-jongho, line 78: **Question: same for the vision config.**

`__init__` here is effectively the same code as `RBLNQwen3VLVisionModelConfig.__init__` — same `deprecate_kwarg`, same int/list normalisation and reverse sort, same `ValueError`. It could be `class RBLNQwen3_5VisionModelConfig(RBLNQwen3VLVisionModelConfig)` with nothing else.

Same question as on the vision model: intentional decoupling, or copy-paste that can be collapsed? Fine to leave if it is the former.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,984 @@
...
+            self.block_tables = torch.arange(self.rbln_config.kvcache_num_blocks, dtype=torch.int16)
+
+    @property
+    def logits_last_dim(self):
+        if self.can_generate():
+            return self.config.text_config.vocab_size
+        else:
+            return self.embedding_dim if hasattr(self, "embedding_dim") else self.config.text_config.hidden_size
+
+    def _create_embedding_layer(self):
```
rebel-jongho, line 613: **Question: the text-side helpers are copies too.**

`_create_embedding_layer`, `_get_position_embeddings`, `_load_submodules` and `logits_last_dim` in this class, plus `can_generate`, `_reconstruct_model_if_needed` and `__post_init__` in `RBLNQwen3_5ForConditionalGeneration`, are byte-identical to their `RBLNQwen3VLModel` / `RBLNQwen3VLForConditionalGeneration` counterparts, and `_preprocess_decoder` is 98% identical. Together with the vision tower this is roughly 250 lines carried over.

The genuinely Qwen3.5-specific overrides — `get_input_info`, `setup_runtime`, `_get_compile_context`, `_preprocess_prefill` — are all clearly different, so a shared base or a mixin would leave those untouched.

Is the full copy intentional? If the answer is that Qwen3.5 should not inherit from a VL-specific class, that is fair and I am happy to leave it.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
```diff
@@ -0,0 +1,961 @@
...
+        super().__init__(model, rbln_config, use_rotary_emb)
+        model.config = original_config
+
+    def get_decoder_layers(self, model: "PreTrainedModel"):
+        return model.get_decoder().layers
+
+    def get_model_layer(self, model: "PreTrainedModel"):
+        return model.get_decoder()
+
+    def prepare_forward_args(self, *args):
```
rebel-jongho, line 850: **Question: why does this override not delegate to `super()`?**

`Qwen3_5_CausalLMWrapper.prepare_forward_args` (line 767) uses the pattern I would expect: pop the Qwen3.5-specific trailing inputs, hand the rest to `super().prepare_forward_args(*args)`, then post-process. This subclass instead reimplements the whole front-popping sequence, and the per-layer linear/full state split is then written twice in the same file — once as a list comprehension at line 784, once as a loop at line 893.

Is there something about the multimodal argument order that blocks delegating to the parent here? If it is just historical, folding the two would remove the risk of the two copies drifting. If there is a real reason, leave it and a one-line comment would help the next reader.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
```diff
@@ -0,0 +1,961 @@
...
+            past_key_values,
+            past_states,
+            position_embeds,
+            conv_state_mask,
+            recurrent_state_mask,
+            valid_mask,
+            batch_idx,
+        )
+
+    def forward(self, *args):
```
rebel-jongho: **Question: is this `forward` override needed at all?**

`Qwen3_5_CausalLMWrapper.forward` (line 789) unpacks the same 16 values in the same order, and the only difference here is that the 12th element is named `position_embeds` instead of `rotary_emb` before being passed as `rotary_emb=`. Since `prepare_forward_args` above already returns `position_embeds` in that slot, deleting this override should be behaviour-preserving and drop 43 lines.

Am I missing a difference? If it is kept for readability, that is fine — just confirm.

rebel-jongho (changes requested):
Reviewed the hybrid Qwen3.5 wiring end to end. The file layout and the class lineage (`RBLNDecoderOnlyModel` -> `...ForCausalLM`, `Wrapper` + `get_rbln_*` factories) match qwen3_vl and gemma4, so the overall shape fits the codebase well, and replacing HF's sequential `for i in range(1, chunk_size)` loop with a single `tri_recur_update` call is exactly the right move.

The comments below fall into two groups.

**Change requests** — the runtime reimplements `RBLNRuntimeModel.prefill_forward` instead of extending it and silently loses three behaviours the base provides (LoRA int-id resolution, prefix caching, `position_ids`), `output_hidden_states` is wired on the model side but never produced by the runtime, and `decode_forward` allocates three tensors per generated token that its own comment says are pruned.

**Questions** — a set of methods copied from qwen3_vl rather than inherited. If the independence is deliberate, say so and leave them as they are; I am asking rather than requesting a change.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: You are right. This hardcode is not necessary

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
rebel-seinpark: Went with the explicit rejection rather than porting the base's guard + mask pre-reveal.

Porting only the attention-mask pre-reveal would fix the full_attention layers but silently leave the linear_attention (GatedDeltaNet) layers wrong: prefix caching would also require carrying each linear layer's conv_state / recurrent_state across the cached prefix, and we have no logic to reconstruct that running state at the prefix boundary. So a mask-only port would look supported while computing the linear layers against an empty state - worse than an honest failure.

So I reject `prefix_cached_len > 0` with `NotImplementedError` (and documented why in a comment) until the hybrid linear-state path can actually be prefix-cached. This keeps the default path (prefill from position 0) unchanged.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
rebel-seinpark: Decided to reject `use_position_ids` for Qwen3.5 (checked against HF `modeling_qwen3_5.py`): neither path needs an explicit `position_ids` graph input.

  - **VL path**: rotary is mRoPE (3D positions for image tokens). HF computes it via `get_rope_index` on the host, and we do the same - the resulting cos/sin is fed to the graph as `position_emb`. The graph never consumes `position_ids`.
  - **Text path**: HF derives `position_ids` from `arange` (i.e. `cache_position`) when none is passed, and mRoPE degenerates to plain RoPE for text-only. Our runtime already uses `cache_position` for the in-graph rotary, so `position_ids` would be redundant.

  So `use_position_ids=True` would only add an unfed graph input. I reject it in `get_input_info` with `NotImplementedError` (VL delegates here, so both are covered), and - since the flag is now always `False` - I removed the dead `use_position_ids` branches (2D-mask layout + `position_ids` input), fixing the attention mask to 4D throughout.

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: Agreed this should be supported, but it's more than a small change here: the custom `_run` keeps only `out[0]` (logits) and returns `hidden_states=None` because the conv/recurrent/KV static caches emit in-place cache-update outputs we skip.

Taking it as a **follow-up**. For now I reject `output_hidden_states=True` at config time (`get_input_info`) with `NotImplementedError` so it fails explicitly instead of hitting the `TypeError` you spotted (VL delegates to this get_input_info, so both are covered).

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: Deliberate. The two towers are decoupled on purpose.

src/optimum/rbln/transformers/models/qwen3_5/configuration_qwen3_5.py
rebel-seinpark: ditto. https://github.com/RBLN-SW/optimum-rbln/pull/631#discussion_r3656038374

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,268 @@
...
+    whatever the previous window/step wrote). ``_run`` maps the named inputs onto the runtime's own
+    (rebel-pruned) input order via ``_index_to_input_name``; the static caches are absent from that order,
+    and the graph's cache-update outputs alias the static addresses (in-place device writes), so only
+    ``logits`` is kept.
+
+    NOTE — multi-window prefill needs FLASH attention. A partial last prefill window
+    (``seq_len % prefill_chunk_size != 0``) attending across PRIOR KV blocks (i.e. the 2nd or later chunk)
+    is mis-lowered by the EAGER attention op (prefill logit pearson ~0.99 vs HF eager; the argmax can flip).
+    FLASH attention (``max_seq_len >= 4096``, ``kvcache_partition_len >= 4096``) computes it correctly
+    (~0.9999). The GatedDeltaNet / linear-state path is accurate under both. Prefer flash for Qwen3.5 hybrid.
```
rebel-kblee: do we need pearsonr comment?

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -0,0 +1,268 @@
...
+        if self.rbln_config.use_lora:
+            named["lora_int_ids"] = lora_int_ids
+
+        # Decode always continues from the real carried state -> ones (no-op). The masks are gated on the
+        # prefill phase in the GatedDeltaNet, so they are pruned from the decode graph and these entries
+        # are simply ignored by the name-based input mapping; passed (precomputed constants) for safety if
+        # they survive. Reused, not reallocated — decode runs once per token.
+        named["conv_state_mask"] = self._conv_mask_ones
+        named["recurrent_state_mask"] = self._recurrent_mask_ones
+        named["valid_mask"] = self._valid_mask_decode
```
rebel-kblee: do we need all these tensors?

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,979 @@
...
+        text_config = model_config.get_text_config()
+        if getattr(text_config, "layer_types", None) is None:
+            raise ValueError("Qwen3.5 requires `layer_types` in the model config.")
+        # Qwen3.5 feeds rotary via the precomputed `position_emb` (cos/sin), so in-graph position_ids are not
+        # part of its design and the runtime never passes them. Reject use_position_ids until it's supported.
+        if rbln_config.use_position_ids:
+            raise NotImplementedError("use_position_ids is not supported for the Qwen3.5 model.")
+        # TODO(seinpark) : output_hidden_states isn't wired yet; planned as a follow-up.
+        if rbln_config.output_hidden_states:
+            raise NotImplementedError("output_hidden_states is not yet supported for the Qwen3.5 model.")
```
rebel-kblee: is it right way to raise not implemented errorr?

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,979 @@
...
+        head_dim = hidden_size // num_heads
+
+        input_infos = []
+        for max_seq_len in rbln_config.max_seq_len:
+            input_info = [
+                ("hidden_states", [max_seq_len, hidden_size], rbln_config.dtype),
+                ("attn_mask", [1, 1, max_seq_len, max_seq_len], rbln_config.dtype),
+                # cos/sin enter the device at fp32 and are cast to the device dtype inside the vision model
+                ("cos", [1, 1, max_seq_len, head_dim], torch.float32),
+                ("sin", [1, 1, max_seq_len, head_dim], torch.float32),
```
rebel-kblee: is this model has no batch configuration?

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,979 @@
...
+        model: Optional["PreTrainedModel"] = None,
+        model_config: "PretrainedConfig" = None,
+        rbln_config: RBLNQwen3_5VisionModelConfig | None = None,
+    ) -> RBLNQwen3_5VisionModelConfig:
+        hidden_size = model_config.hidden_size
+        num_heads = model_config.num_heads
+        head_dim = hidden_size // num_heads
+
+        input_infos = []
+        for max_seq_len in rbln_config.max_seq_len:
```
rebel-kblee, line 384: can you explain about bucketing ?

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
```diff
@@ -0,0 +1,961 @@
...
+        base = list(super().prepare_forward_args(*args))
+        # base[-2] is the combined per-layer state list (2/layer, indexed by layer_idx): split by layer type
+        # into past_key_values (full-attn KV) and past_states (linear conv/recurrent), both full-length with
+        # None at the other type's indices so [layer_idx] indexing still works.
+        linear = {i for i, t in enumerate(self.config.layer_types) if t == "linear_attention"}
+        combined = base[-2]
+        base[-2] = [pair if i not in linear else None for i, pair in enumerate(combined)]  # past_key_values
+        past_states = [pair if i in linear else None for i, pair in enumerate(combined)]
+        base.insert(-1, past_states)  # insert past_states just before rotary_emb (the last element)
+        return (*base, conv_state_mask, recurrent_state_mask, valid_mask, batch_idx)
```
rebel-kblee: I think it's fragile code for indexing the argument. Is there any format?

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: Same as the vision model/config

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
rebel-seinpark: You are right. I fixed it

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: Good catch! followed the base's branch shape.

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
rebel-seinpark: They(`conv_state_mask`, `recur_state_mask`, `valid_mask`) are not needed in decode as you said. Thanks

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: Could you take a look at this commit?
https://github.com/RBLN-SW/optimum-rbln/pull/631/changes/9362a642017c672420bc02c1e02445bda8a5b2aa

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
rebel-seinpark: The positional coupling here comes from the base `DecoderOnlyWrapper.prepare_forward_args` protocol (it returns/consumes a positional tuple across the whole decoder family), so it's hard to remove entirely without diverging from that. :(

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: How about [this way](https://github.com/RBLN-SW/optimum-rbln/pull/631/changes/f7d99381ea23eabd8412950d9718060743c747f3)?

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,954 @@
...
+                    kvcache_metas.append(
+                        KVCacheMeta(
+                            name=f"recurrent_state_{layer_idx}",
+                            layer_index=layer_idx,
+                            shape=recurrent_shape,
+                            layer_type="linear_attention",
+                            is_auto=False,
+                            dtype=_state_dtype,
+                        )
+                    )
```
rebel-jongho, line 246: do not create kvcachemeta directly. use `.make`

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -0,0 +1,954 @@
...
+            input_info.extend([(meta.name, meta.compile_shape, meta.dtype) for meta in rbln_config.kvcache_metas])
+        else:
+            kvcache_dtype = rbln_config.dtype
+            if rbln_config.quantization and rbln_config.quantization.kv_caches == "fp8":
+                kvcache_dtype = "float8_e4m3fn"
+
+            # conv/recurrent caches are one shared static tensor, so sized to the max batch (not batch_size=1 for
+            # prefill). Prefill writes its own slot (batch_idx); decode runs the full batch.
+            state_batch = rbln_config.batch_size
+            _state_dtype = RBLNCompileConfig.normalize_dtype(rbln_config.dtype)
```
rebel-jongho, line 223: why use this temporal variable?

rebel-jongho (changes requested):

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: just code readability. It doesn't matter if you want to modify this temp var.

rebel-kblee (approved):
👍

src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
rebel-seinpark: TODO: refactor KVCacheMeta

rebel-jongho (approved):

PR #634, opened 2026-07-16
model: NPU-aware prefill_chunk_size default (512 on RBLN-CR)

rebel-thkim:
## What

Make the default `prefill_chunk_size` depend on the NPU type — **512** on **RBLN-CR** devices and **128** otherwise. CR NPUs use a larger prefill chunk for better prefill performance (PCC).

## Changes

- **`set_default_values`**: Resolve the NPU-based `prefill_chunk_size` default and its divisible-by-64 validation here, using `rebel.get_npu_name(0)` (guarded with `or ""` so compiling on a host without an attached NPU falls back to 128 instead of raising on `"RBLN-CR" in None`). This is the shared entry point that both decoderonly and gemma4 already route through via `_update_attention_config`, so the default is defined in one place.
- **decoderonly**: Remove the `prefill_chunk_size` default/validation from the config constructor; `_update_attention_config` now captures the value returned by `set_default_values`.
- **gemma4**: Drop the hardcoded 128 default from `RBLNGemma4ForCausalLMConfig` so it resolves at compile time through the shared path.

## Notes

- gemma3 (`prefill_chunk_size or 256`) is **left as-is** — existing behavior preserved.

## Review feedback addressed

- Dropped the `_get_npu_name` helper and its `RBLN_FORCE_NPU_NAME` / `RBLN_TARGET_SOC` env fallbacks; `_resolve_npu` / `warn_deprecated_npu` are back to using `rebel.get_npu_name` directly.
- Moved the `prefill_chunk_size` default into `set_default_values` instead of duplicating it across each `_update_rbln_config`.

## Test

- `py_compile` + `ruff check` pass.
- Added unit tests (`tests/test_config.py::TestPrefillChunkSizeDefault`) covering NPU-aware default resolution (512 on RBLN-CR, 128 otherwise), attached-NPU fallback, no-NPU fallback, explicit-value precedence, and divisible-by-64 validation — 8 passed.
- Added an e2e wiring test (`test_prefill_chunk_size_npu_wiring_e2e`): pins `npu=RBLN-CR03` in `rbln_config`, compiles tiny-llama without a CR device, asserts the resolved `prefill_chunk_size` is 512 and survives save/reload. Verified locally (9 passed, ~13s).


src/optimum/rbln/utils/runtime_utils.py
```diff
@@ -45,11 +46,18 @@ def is_compiler_supports_chiplet_alloc() -> bool:
     return hasattr(rebel.RBLNCompiledModel, "get_alloc_per_chiplet_by_key")
 
 
+def _get_npu_name() -> str:
+    device_name = (
+        rebel.get_npu_name(0) or os.environ.get("RBLN_FORCE_NPU_NAME") or os.environ.get("RBLN_TARGET_SOC") or None
+    )
```
rebel-jongho: This is not needed. just use rebel.get_npu_name()

rebel-jongho (commented):
how about modifying `set_default_values` to set the default value of pcs?

rebel-jongho (approved):
LGTM

do we need any tests? or it's just a simple functionality?

rebel-thkim:
Resolved the merge conflict with `dev` (the repo-wide `Optional[X]` → `X | None` typing migration touched the same lines) and added unit tests in `tests/test_config.py` covering the NPU-aware `prefill_chunk_size` resolution: 512 on RBLN-CR, 128 otherwise, fallback to the attached NPU (and to 128 when no NPU is attached), explicit values taking precedence, and the divisible-by-64 validation. All 8 tests pass locally without device compilation.

PR #637, opened 2026-07-21
performance: add memory_budget for kv-cache block estimation

rebel-jongho:
## Type of Change
- [x] Performance improvement

## Changes Overview
Adds `memory_budget` to `RBLNDecoderOnlyModelConfig` to bound the device DRAM the automatic `kvcache_num_blocks` estimation may assume.
- Prefer a float in (0, 1] as a fraction of the NPU available DRAM (e.g. `0.8` for 80%), matching vllm's `gpu_memory_utilization`.
- A `"80%"` string is also accepted (handy for passing through the compile CLI), as are bytes given as an int or string (`"10GB"`, `"512MB"`).
- The fraction is of the available DRAM (after the per-chiplet system reserve), not the raw total. Defaults to None (full available DRAM) and must not exceed it (`ValueError` otherwise).

Also adds `RBLNDecoderOnlyFlashAttentionMixin._required_memory_at(compiled_models, rbln_config, num_blocks)`, returning the device-wide kv-cache DRAM at a given block count with 2MB alignment applied (non-linear in the block count). It works for any current block size — resizable tensors are normalized to per-block bytes using `rbln_config.kvcache_num_blocks`. The aligned per-chiplet math is factored into `_kvcache_bytes_per_chiplet`, shared with the block search (unchanged behavior).

```bash
optimum-rbln-cli --model-id meta-llama/Llama-2-7b-chat-hf --memory-budget 0.8
optimum-rbln-cli --model-id meta-llama/Llama-2-7b-chat-hf --memory-budget 80%
optimum-rbln-cli --model-id meta-llama/Llama-2-7b-chat-hf --memory-budget 100GB
```

## Motivation and Context
Users need to constrain kv-cache block estimation to a memory budget smaller than the full device (e.g. to leave headroom), expressed intuitively as a fraction or in bytes.

Note: the float-vs-`%` representation follows reviewer feedback — float is the recommended form (per vllm), with `"%"` kept for CLI convenience.

## How to test
- Estimation with a smaller budget yields fewer blocks (monotonic); `None` reproduces the current result; a budget above the NPU available DRAM raises `ValueError`.
- `_required_memory_at(N)` is monotonic in `N`, equals the current footprint at `N = kvcache_num_blocks`, and stays correct for an already-resized artifact.

### DIFF
--- a/src/optimum/rbln/transformers/modeling_attention_utils.py
+++ b/src/optimum/rbln/transformers/modeling_attention_utils.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, Optional, Tuple
 import rebel
 
 from ..utils.logging import get_logger
-from ..utils.runtime_utils import get_available_dram_per_chiplet
+from ..utils.runtime_utils import get_available_dram_per_chiplet, parse_byte_size
 
 
 if TYPE_CHECKING:
@@ -168,6 +168,33 @@ def format_byte_size(nbytes: int) -> str:
         return f"{nbytes / 1024**3:.2f} GB"
 
 
+def _resolve_memory_budget(memory_budget: Optional[object], available_total: int) -> int:
+    """Resolve `memory_budget` to usable DRAM bytes (system reserve excluded), capped at available_total.
+
+    None -> available_total; a float in (0, 1] (or a "80%" string) -> that fraction of it;
+    int/"10GB"/"512MB" -> parsed bytes. `available_total` is the device-wide available DRAM after
+    the per-chiplet system reserve.
+    """
+    if memory_budget is None:
+        return available_total
+    fraction = None
+    if isinstance(memory_budget, float):
+        fraction = memory_budget
+    elif isinstance(memory_budget, str) and memory_budget.strip().endswith("%"):
+        fraction = float(memory_budget.strip()[:-1]) / 100
+    if fraction is not None:
+        if not 0.0 < fraction <= 1.0:
+            raise ValueError(f"memory_budget fraction must be in (0, 1] (or (0%, 100%]), got {memory_budget!r}.")
+        budget = int(available_total * fraction)
+    else:
+        budget = parse_byte_size(memory_budget)
+    if budget > available_total:
+        raise ValueError(
+            f"memory_budget ({budget} bytes) exceeds the target NPU's available DRAM ({available_total} bytes)."
+        )
+    return budget
+
+
 class RBLNDecoderOnlyFlashAttentionMixin:
     @classmethod
     def set_kvcache_num_blocks_after_compilation(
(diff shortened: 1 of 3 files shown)


src/optimum/rbln/transformers/models/decoderonly/configuration_decoderonly.py
```diff
@@ -97,6 +98,12 @@ def __init__(
             kvcache_num_blocks (Optional[int]): The total number of blocks to allocate for the
                 PagedAttention KV cache at compile time. Defaults to 0 (automatically determined).
                 See the "KV Cache Number of Blocks (`kvcache_num_blocks`)" section below for details.
+            memory_budget (Optional[Union[int, float, str]]): Usable DRAM budget (system reserve
+                excluded) used when auto-estimating `kvcache_num_blocks`. Prefer a float in (0, 1]
+                as a fraction of the NPU available DRAM (e.g. 0.8 for 80%). A "80%" string is also
+                accepted (convenient for passing through the compile CLI), as are bytes given as an
```
rebel-jonghewk, line 104: nit: `convenient for passing through the compile CLI` <- add it after compile ci feature

src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -325,6 +341,47 @@ def check_memory_fits(multiplier: int) -> Tuple[bool, dict[Tuple[int, int], int]
...
+        """Total device-wide kv-cache DRAM (bytes) at `num_blocks`, with 2MB alignment applied.
+
+        Works for any current block size: the buffers' block count is `rbln_config.kvcache_num_blocks`
+        (0 for the unresized compile baseline is treated as 1). Bytes are not linear in `num_blocks`
+        because alignment is applied after scaling.
+        """
+        kvcache_tensor_sizes = compiled_models["prefill"].exp_get_dram_tensor_sizes()
+        current_blocks = rbln_config.kvcache_num_blocks or 1
+        return sum(
+            cls._kvcache_bytes_per_chiplet(kvcache_tensor_sizes, rbln_config, num_blocks, current_blocks).values()
```
rebel-jonghewk, line 382: When does this get called?

rebel-jonghewk (approved):

src/optimum/rbln/transformers/models/decoderonly/configuration_decoderonly.py
rebel-jongho: The CLI already exists today, and `memory_budget` can be used at CLI compile time right now (e.g. `optimum-rbln-cli --model-id ... --memory-budget 0.8`). I think this may have read as a feature needed by #638, but that's not the case — it works with the current compile CLI and doesn't depend on #638. So the note refers to the existing compile CLI.

src/optimum/rbln/transformers/modeling_attention_utils.py
rebel-jongho: This is intentional — `_required_memory_at` is added as a query helper for external tooling (to inspect the device-wide kv-cache DRAM at a given block count). It isn't called internally in this PR.

PR #614, opened 2026-06-27
other(deprecation): apply cutoff to pre-, post-, and dev-releases

rebel-junamsong:
## Summary

`@deprecate_method` / `@deprecate_kwarg` raise after `__version__ >= cutoff`. Because `packaging.version` orders `X.Y.Za1 < X.Y.Z < X.Y.Z.post1`, **pre-/post-/dev-releases never trigger the raise** — deprecated calls only blow up at the final tag.

This PR normalizes the comparison through `Version(...).base_version` so any release on the same `X.Y.Z` line — `dev`, `aN`, `bN`, `rcN`, `final`, `.postN` — hits the cutoff together.

## Behavior change

| current | before | after |
|---|---|---|
| `X.Y.Z.dev0` / `aN` / `bN` / `rcN` | warn | **raise** |
| `X.Y.Z` | raise | raise |
| `X.Y.Z.postN` / `X.Y.(Z+1)` | raise | raise |
| `< X.Y.Z` | warn | warn (unchanged) |

## Tests

- `tests/test_base.py::test_deprecate_method_raises_at_or_past_cutoff` — parametrized integration test that mocks `__version__`, decorates a stub with `@deprecate_method(version="1.0.0")`, and asserts the raise outcome across 9 PEP 440 suffix variants.
- `tests/test_config.py::test_load_config_object` — switched from `.load(...)` (deprecated) to `.from_pretrained(...)`. The test no longer triggers the raise it's now subject to.

## Related

[235075e4](https://github.com/RBLN-SW/optimum-rbln/commit/235075e4e9991bcdae9a89897bf0f0a2f3c6d92f) — a stale `.load()` call that survived all `0.11.0a*` builds and only raised at the `0.11.0` tag. Exactly the pattern this PR catches earlier.

PR #644, opened 2026-07-27
refactor(config): remove RBLNConfig.load() past its 0.11.0 removal deadline

rebel-jongho:
# Pull Request Description

## Type of Change
- [x] Code refactoring

## Changes Overview

Deletes the two deprecated `load()` classmethods from `src/optimum/rbln/configuration_utils.py`, decorator included.

- `RBLNAutoConfig.load()` and `RBLNModelConfig.load()` were thin wrappers forwarding their arguments verbatim to `from_pretrained()`, so callers only need to change the method name.
- The `deprecate_method` import goes with them, since nothing in that module uses it anymore. The helper itself stays in `utils/deprecation.py` next to `deprecate_kwarg` and remains covered by `tests/test_base.py::test_deprecate_method_raises_at_or_past_cutoff`.

No test changes are needed: `test_load_config_object` was already switched to `from_pretrained()` in 48217ce0.

## Motivation and Context

Both methods carried `@deprecate_method(version="0.11.0", new_method="from_pretrained")`, and every 0.11.x build is now past that cutoff, so calling either one raises `ValueError` instead of warning. Two things put us there:

1. `v0.11.0.post1` became reachable from `dev` with the back-merge in #618, which pushed the hatch-vcs derived version to `0.11.0.post2.devN`.
2. #614 normalized the cutoff to compare PEP 440 base versions, so alpha and dev builds (`0.11.1aN.devM`) are past it as well.

The deprecation policy table at the top of `utils/deprecation.py` prescribes exactly this sequence: raise for one version, then drop the decorator. This PR performs that removal step.

No callers remain. `auto_factory.py` moved to `from_pretrained()` in 235075e4, `test_load_config_object` in 48217ce0, and a repository-wide search turns up no other references beyond the unrelated `load_from_dict`.

## Verification

`ruff check` and `ruff format` pass on the changed file. The change is a pure deletion of two unreachable wrappers plus the import that served them, so behaviour is unchanged for every code path that still works today.

## Out of Scope

The three `max_seq_lens` → `max_seq_len` `deprecate_kwarg` entries on the Qwen2-VL / Qwen2.5-VL / Qwen3-VL vision configs share the same 0.11.0 cutoff and are equally past it, but no test or internal caller exercises them, so they are left for a follow-up.

## Related Issues

SR-365

### DIFF
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -24,7 +24,7 @@ import torch
 from packaging.version import Version
 
 from .__version__ import __version__
-from .utils.deprecation import deprecate_kwarg, deprecate_method, warn_deprecated_npu
+from .utils.deprecation import deprecate_kwarg, warn_deprecated_npu
 from .utils.logging import get_logger
 from .utils.runtime_utils import ContextRblnConfig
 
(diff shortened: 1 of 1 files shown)


rebel-thkim:
already merged in https://github.com/RBLN-SW/optimum-rbln/pull/645

PR #646, opened 2026-07-27
model: fix default prefill_chunk_size to max_seq_len for bidirectional prefill (paligemma)

rebel-thkim:
## Problem

PaliGemma's language model (and ColPali's, through its `vlm` submodule) compiles prefill with **bidirectional attention** — the attention op receives `is_bidirectional=True` when the config has `use_attention_mask=True` + `use_position_ids=True` and no `image_prefill` phase. Bidirectional attention only spans a single prefill chunk (earlier chunks' KV are frozen), so the whole prompt must fit in one chunk: effectively `prefill_chunk_size == max_seq_len`.

Nothing enforced this. With `prefill_chunk_size` unset, the generic NPU default (128, or 512 on RBLN-CR) applied — even a single 224px image (256 soft tokens) then spans two chunks and produces **silently wrong logits with no error**. The only guidance was the `prefill_chunk_size: 8192` pin in docstring examples.

## Why the fix is keyed on config flags, not the model class

The same language-model class is causal or bidirectional depending on which parent instantiates it: standalone Gemma is causal; under Gemma3/4 the text prefill is causal (they route vision runs to a separate bidirectional `image_prefill` graph); under PaliGemma it is bidirectional. The parent already communicates that semantics through the flags it injects into the submodule config, so the constraint hangs on the exact predicate the attention op uses (no new config field is introduced):

```python
use_attention_mask and use_position_ids and not use_image_prefill
```

## Changes

- **`configuration_decoderonly.py`** — new `use_bidirectional_prefill` property: the single definition of the flag combo that maps to `is_bidirectional=True` in the compiled attention op.
- **`modeling_decoderonly.py` `_update_attention_config`** — when the combo is active and `prefill_chunk_size` is unset, default it to `max_seq_len` instead of the generic NPU default. Lives on `RBLNDecoderOnlyModel`, so both the `ForCausalLM` (PaliGemma generation) and `Model` (ColPali retrieval) paths are covered. Gemma3/4 are unaffected (`use_image_prefill=True`).
- **`decoderonly_runtime_utils.py` `prefill_forward`** — if the input still exceeds `prefill_chunk_size` in bidirectional mode (e.g. a user explicitly compiled a smaller chunk), raise a clear `ValueError` instead of silently chunking into wrong outputs.
- **Docstrings** — drop the now-redundant manual `prefill_chunk_size: 8192` pins from the PaliGemma / ColPali examples. 8192 is PaliGemma's `max_position_embeddings`, i.e. exactly what the new default resolves to, so compiled artifacts for the documented usage are unchanged. (The ColPali example also had a missing-comma syntax error that goes away with the block.)

## Behavior changes

- PaliGemma/ColPali compiled **without** an explicit `prefill_chunk_size`: chunk goes from 128/512 (broken outputs) to `max_seq_len` (correct). Memory/compile-time increases accordingly — this is the cost of correctness, and matches what the docstrings already recommended.
- PaliGemma/ColPali compiled **with** the documented 8192 pin: identical artifact as before.
- A bidirectional model fed an input longer than its compiled chunk now fails loudly at runtime instead of returning wrong logits.
- Causal models, Gemma3/4, Qwen-VL, ColQwen2: no change.

Related: RBLN-SW/vllm-rbln#834 derives the same value (`prefill_chunk_size = max_model_len`) on the vLLM compile path; with this PR, even a caller that omits it gets the correct chunk from optimum-rbln itself.

## Verification

- `ruff format --check` / `ruff check --select I` / `py_compile` pass on all edited files.
- Unit tests could not run in my local environment (pre-existing `rebel._C` version mismatch breaks collection); please rely on CI.

### DIFF
--- a/src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
+++ b/src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
@@ -484,6 +484,12 @@ class RBLNDecoderOnlyModel(RBLNModel, RBLNDecoderOnlyFlashAttentionMixin):
     def _update_attention_config(
         cls, model: PreTrainedModel, model_config: PretrainedConfig, rbln_config: RBLNDecoderOnlyModelForCausalLMConfig
     ):
+        # Bidirectional prefill attends across the whole prompt, so the prompt must fit in a
+        # single prefill chunk — default the chunk size to max_seq_len instead of the generic
+        # NPU default, which would silently truncate the bidirectional context.
+        if rbln_config.prefill_chunk_size is None and rbln_config.use_bidirectional_prefill:
+            rbln_config.prefill_chunk_size = rbln_config.max_seq_len
+
         (
             rbln_config.attn_impl,
             rbln_config.kvcache_partition_len,
--- a/src/optimum/rbln/transformers/models/colpali/configuration_colpali.py
+++ b/src/optimum/rbln/transformers/models/colpali/configuration_colpali.py
@@ -33,11 +33,8 @@ class RBLNColPaliForRetrievalConfig(RBLNModelConfig):
 
         # Create a configuration object
         config = RBLNColPaliForRetrievalConfig(
-            vlm={
-                "language_model": {"prefill_chunk_size": 8192},
-            }
             output_hidden_states=False,
-            num_devices=4
+            num_devices=4,
         )
 
         # Use the configuration with from_pretrained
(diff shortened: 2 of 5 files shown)


src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
```diff
@@ -484,6 +484,12 @@ def _update_sliding_window_config(
     def _update_attention_config(
         cls, model: PreTrainedModel, model_config: PretrainedConfig, rbln_config: RBLNDecoderOnlyModelForCausalLMConfig
     ):
+        # Bidirectional prefill attends across the whole prompt, so the prompt must fit in a
+        # single prefill chunk — default the chunk size to max_seq_len instead of the generic
+        # NPU default, which would silently truncate the bidirectional context.
+        if rbln_config.prefill_chunk_size is None and rbln_config.use_bidirectional_prefill:
+            rbln_config.prefill_chunk_size = rbln_config.max_seq_len
+
```
rebel-kblee, line 492: is it correct to override to max_seq_len?

rebel-kblee (approved):
👍

PR #647, opened 2026-07-28
model:  fix force vision_tower output_hidden_states for Llava

rebel-thkim:
## Problem

`RBLNLlavaForConditionalGeneration` extracts image features from a vision hidden layer (`vision_feature_layer`) and always calls the vision tower with `output_hidden_states=True` at inference. But nothing forces the **compile-time** value:

1. Llava's config does not set it for the `vision_tower` submodule, so the CLIP/Pixtral vision config falls back to the HF model config's `output_hidden_states` — **False** for llava-hf checkpoints.
2. The tower then compiles without hidden states, and the first inference fails with `ValueError: Variable output_hidden_states True is not equal to rbln_config.output_hidden_states False. Please compile again with the correct argument.`

So a user who does not manually pass `vision_tower.output_hidden_states=True` (it was only shown in a docstring example) gets an artifact that compiles fine and breaks at first use.

## Fix

Apply the same forced kwarg **LlavaNext already uses** (`configuration_llava_next.py`) to Llava's `vision_tower` submodule init:

```python
self.vision_tower = self.initialize_submodule_config(
    submodule_config=vision_tower,
    batch_size=1,
    output_hidden_states=True,  # Llava requires output_hidden_states to be True
    force_kwargs=True,
)
```

- A dict/None `vision_tower` config now always compiles with hidden states; an explicit conflicting `False` raises a parameter-conflict error at config time instead of failing at inference.
- The now-redundant manual pin is dropped from the `from_pretrained` docstring example (matching LlavaNext's example).
- The typed-config example (`RBLNCLIPVisionModelConfig(..., output_hidden_states=True)`) keeps the explicit value: a config **instance** bypasses forced kwargs in `initialize_submodule_config`, so it is still required on that path.
- Pixtral vision towers are covered too — `RBLNPixtralVisionModelConfig` accepts the same kwarg.

## Verification

- `ruff format --check` / `ruff check --select I` / `py_compile` pass.
- Unit tests could not run locally (pre-existing `rebel._C` version mismatch); please rely on CI.


rebel-thkim:
https://github.com/huggingface/transformers/blob/94f09cfec149050b5355bab7f207ac69e21f1a02/src/transformers/models/llava/modeling_llava.py#L156

rebel-kblee (approved):
👍

PR #648, opened 2026-07-28
model: restore batch_size floor in num_min_blocks

rebel-jongho:
# Pull Request Description

## Type of Change

## Changes Overview

`num_min_blocks` now keeps `batch_size` as a floor under flash attention:

```python
if self.attn_impl == "flash_attn":
    blocks_in_use = max(self.max_seq_len // self.kvcache_block_size, self.batch_size)
    return min(blocks_in_use + 1, self.num_full_blocks)
```

The pool must hold either one full-length sequence or one block per batch row, plus the spare block `replace_empty_block` hands out for empty block-table slots.

Examples at `kvcache_block_size = 16384`, `num_min_blocks` before → after:

| batch_size | max_seq_len | num_full_blocks | num_min_blocks |
| --- | --- | --- | --- |
| 4 | 131072 | 32 | 9 → 9 |
| 16 | 131072 | 128 | 9 → 17 |
| 64 | 32768 | 128 | 3 → 65 |

The rejection message now reports the estimated and required counts and names `batch_size`, and the `kvcache_num_blocks` docstring is corrected to match.

## Motivation and Context

Before #437 the flash attention path also required `kvcache_num_blocks >= batch_size`. Folding the checks into `num_min_blocks` dropped that term, so a configuration with fewer blocks than batch rows compiled and then failed in `generate()` with `No memory blocks are available for allocation`, since `RBLNPageTableManager` assigns a distinct block per row. It is now rejected at compile time.

## Related Issues

None.

----

### DIFF
--- a/src/optimum/rbln/transformers/modeling_attention_utils.py
+++ b/src/optimum/rbln/transformers/modeling_attention_utils.py
@@ -224,8 +224,9 @@ class RBLNDecoderOnlyFlashAttentionMixin:
         )
         if rbln_config.kvcache_num_blocks < rbln_config.num_min_blocks:
             raise ValueError(
-                "Memory is not enough for full sequence length. "
-                "Please consider decreasing `max_seq_len` to reduce the number of blocks."
+                f"Memory is not enough for the minimum number of kv-cache blocks "
+                f"({rbln_config.kvcache_num_blocks} < {rbln_config.num_min_blocks}). Please consider decreasing "
+                f"`max_seq_len` or `batch_size` ({rbln_config.batch_size}) to reduce the number of blocks."
             )
         cls.multiply_kv_cache_num_blocks(
             compiled_models=compiled_models, rbln_config=rbln_config, multiplier=rbln_config.kvcache_num_blocks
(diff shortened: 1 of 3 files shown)


src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -212,8 +224,9 @@ def _log_memory_usage(compiled_models: dict[str, rebel.RBLNCompiledModel], prefi
         )
         if rbln_config.kvcache_num_blocks < rbln_config.num_min_blocks:
             raise ValueError(
-                "Memory is not enough for full sequence length. "
-                "Please consider decreasing `max_seq_len` to reduce the number of blocks."
+                f"Memory is not enough for the minimum number of kv-cache blocks "
+                f"({rbln_config.kvcache_num_blocks} < {rbln_config.num_min_blocks}). Please consider decreasing "
+                f"`max_seq_len` or `batch_size` ({rbln_config.batch_size}) to reduce the number of blocks."
             )
```
rebel-eunji, line 230: 'Memory is not enough','minimum number of' sounds a bit awkward. How about we take this opportunity to rephrase it?
```python
raise ValueError(
    f"Insufficient memory for the required KV cache: "
    f"only {rbln_config.kvcache_num_blocks} blocks can be allocated, "
    f"but at least {rbln_config.num_min_blocks} blocks are required. "
    f"Consider reducing max_seq_len or batch_size "
    f"(currently {rbln_config.batch_size}) to lower the memory requirement."
)
```

rebel-eunji (approved):

rebel-thkim (approved):
👍

PR #650, opened 2026-07-29
model: gemma4 stop propagating batch_size to the language_model submodule

rebel-thkim:
`RBLNGemma4ForConditionalGenerationConfig` rejects a batch size the caller explicitly asked for.

```
ValueError: Parameter conflict for 'batch_size':
            submodule_config has 4, but kwargs has 1. Using kwargs value: 1
```

The config defaults its own `batch_size` to 1, then forces that onto the language model:

```python
self.batch_size = batch_size or 1
...
self.language_model = self.initialize_submodule_config(
    submodule_config=language_model,
    batch_size=self.batch_size,     # 1, because nothing set it
    force_kwargs=True,              # -> any mismatch raises
    use_inputs_embeds=True,
)
```

A caller that configures only the submodule never sets the top-level key, so the parent silently becomes 1 and conflicts with the submodule's 4. That is what vllm-rbln builds for every composite VLM — `{"language_model": {"batch_size": N, ...}}`, no top-level key.

## Why dropping it is safe

Nothing reads the parent's copy. Every batch-dependent shape comes off a submodule config: `RBLNGemma4VisionModel` uses the vision tower's (pinned to 1), `RBLNGemma4ForCausalLM` uses the language model's. `RBLNGemma4ForConditionalGeneration` has no `rbln_config.batch_size` reference.

gemma4 is also the only composite config that does this:

| config | propagates `batch_size` to `language_model` | on mismatch |
|---|---|---|
| gemma3, llava, llava_next, blip_2 | no | submodule wins |
| paligemma | yes, without `force_kwargs` | submodule wins |
| **gemma4** | yes, with `force_kwargs=True` | **raises** |

`force_kwargs=True` stays; it still guards `use_inputs_embeds`, which gemma4 requires.

## Verified

Against the installed `0.11.1a4` and the patched tree, with the rbln_config vllm-rbln builds for `-b 4`:

| case | before | after |
|---|---|---|
| `language_model={"batch_size": 4}`, parent unset | `ValueError` | ok — lm 4, vision 1 |
| `language_model={"use_inputs_embeds": False}` | raises | raises |
| vision tower | forced to 1 | forced to 1 |

## Also

Dropped `Using kwargs value: {value}` from the conflict message — it raises rather than using anything, and the claim sends you looking in the wrong place. Message only.

Reported by `standalone-rebel-compiler-vllm-ci-v2` build 9, `compile gemma4-moe-b4-l6`. Any `-b N` with N != 1 hits it.


rebel-kblee (approved):
👍

rebel-seinpark (approved):

PR #655, opened 2026-07-30
model: precompute Qwen3.5 ViT rotary cos/sin tables once at load

rebel-seinpark:
## Summary
The Qwen3.5 vision `rot_pos_emb` rebuilt the rotary freq table and computed `cos()/sin()` on **every forward**, over a tensor sized to the current input's `max(h, w)`. That per-call sin/cos over a variable-length tensor can differ in the last bits depending on the CPU's OMP/MKL thread count and tensor size.

## Change
- Precompute the cos/sin tables **once in `__post_init__`**, up to the largest ViT bucket (`max(h,w) <= h*w <= largest bucket`, so every valid input is covered).
- `rot_pos_emb` now **gathers** cos/sin per token from those fixed-size tables instead of recomputing.
- The tables live on the host (CPU) and are fed to the compiled vision graph as `cos`/`sin` inputs, same as before.

## Why it's equivalent
Positional freqs are absolute (`freq_table[p] = p * inv_freq`) and `cos` is elementwise, so `freq_table.cos()[pos_ids] == freq_table[pos_ids].cos()`. The gathered values are identical to the previous per-forward path — only now computed once on a fixed-size tensor, so they're reproducible across runs / thread counts.

## Verification
- `tests/test_llm.py::TestQwen3_5ForCausalLM` + `TestQwen3_5ForConditionalGeneration`: **15 passed, 1 skipped, 6 subtests passed** (ViT correlation preserved).

### DIFF
--- a/src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
+++ b/src/optimum/rbln/transformers/models/qwen3_5/modeling_qwen3_5.py
@@ -431,7 +431,7 @@ class RBLNQwen3_5VisionModel(RBLNModel):
             pos_ids[offset : offset + num_tokens] = coords
             offset += num_tokens
 
-        # Gather cos/sin from the tables precomputed at object creation (fixed size -> deterministic).
+        # Gather cos/sin from the tables precomputed at object creation
         cos = self.rotary_cos_table[pos_ids].flatten(1)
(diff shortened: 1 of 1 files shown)


rebel-thkim (approved):
👍

rebel-jongho (approved):

Issue #649, opened 2026-07-28
Empty last_hidden_state when prompt length is a multiple of prefill_chunk_size


rebel-sbpark:
## What happens

`RBLNQwen3Model` returns an empty `last_hidden_state` whenever the prompt length is an exact
multiple of `prefill_chunk_size`.

`Qwen/Qwen3-Embedding-0.6B`, `max_seq_len=2048`, `prefill_chunk_size=128`:

| tokens | L % 128 | `last_hidden_state.shape` |
|---|---|---|
| 127 | 127 | `(1, 127, 1024)` |
| **128** | **0** | **`(1, 0, 1024)`** |
| 129 | 1 | `(1, 129, 1024)` |
| 255 | 127 | `(1, 255, 1024)` |
| **256** | **0** | **`(1, 0, 1024)`** |
| **384 / 512 / 1024** | **0** | **`(1, 0, 1024)`** |

Downstream code that takes the last token (`h[:, -1, :]`) then raises
`IndexError: index -1 is out of bounds for dimension 1 with size 0`. There is no warning, and
it only happens for certain input lengths, so it is easy to miss.

```python
from optimum.rbln import RBLNQwen3Model
import torch

m = RBLNQwen3Model.from_pretrained("Qwen/Qwen3-Embedding-0.6B", export=True,
                                   rbln_batch_size=1, rbln_max_seq_len=2048)
for L in (127, 128, 129):
    ids = torch.ones((1, L), dtype=torch.int64)
    out = m(input_ids=ids, attention_mask=torch.ones_like(ids))
    print(L, out.last_hidden_state.shape)
```

## Why

`decoderonly_runtime_utils.py` (line numbers from `main`):

```python
620:  # Aggregate output_logits
621:  padding_size = (self.rbln_config.prefill_chunk_size - query_length) % self.rbln_config.prefill_chunk_size
622:  if self.rbln_config.logits_to_keep == 1:
623:      output_logits = output_logits
624:  elif self.rbln_config.logits_to_keep > 1:
625:      output_logits = output_logits[:, -padding_size - self.rbln_config.logits_to_keep : -padding_size, :]
626:  else:
627:      output_logits = output_logits[:, :-padding_size, :]
628:
629:  all_hidden_states = None
630:  if self.rbln_config.output_hidden_states:
631:      all_hidden_states = [
632:          output_hidden_state[:, :-padding_size, :] for output_hidden_state in output_hidden_states
633:      ]
```

[...]

PR #659, opened 2026-07-30
model(qwen3_vl): fix prefill output aggregation for padded batches

### LINKED ISSUE #658: Qwen3-VL prefill output aggregation: padded-batch hidden states crash, left-padded mask corruption, and out_buffers index shadowing
Follow-up to #649: auditing the other prefill output-aggregation paths for the same family of length-bookkeeping bugs turned up three defects in `RBLNQwen3VLRuntimeModel.prefill_forward` (`src/optimum/rbln/transformers/models/qwen3_vl/qwen3_vl_runtime_utils.py`). Line numbers from current `dev`.

## 1. `output_hidden_states=True` crashes on any padded batch

The runtime head-trims hidden states to the valid length:

```python
285:  output_hidden_states = tuple(hs[:, :query_length, :] for hs in output_hidden_states)
```

but every model-level aggregation copies them into full-mask-width buffers:

```python
# modeling_qwen3_vl.py (RBLNQwen3VLModel.forward and RBLNQwen3VLForConditionalGeneration.forward)
758:   all_hidden_states[l_idx][b_idx].copy_(output.hidden_states[l_idx][0])
1014:  all_hidden_states[l_idx][b_idx].copy_(output.hidden_states[l_idx][0])
```

As soon as any batch row is padded (`query_length < seq_len`), `copy_` raises a shape-mismatch RuntimeError. Every other model family is consistent: base decoder-only / qwen2_vl / exaone4_5 return mask-width hidden states and `copy_`; gemma4 returns valid-length and re-scatters with `index_copy_(mask_indices)`. Only qwen3_vl mixes the two conventions.

## 2. Left-padded masks return wrong values (silent)

`_prepare_prefill_outputs` offsets the chunk out-buffer views by the first non-masked position (`valid_start_index`), so valid rows live at `[valid_start, valid_start + query_length)` of the buffer. The aggregation reads the buffer head instead:

```python
264/266:  output_logits = output_logits[:, :query_length, :]
285:      hs[:, :query_length, :]
```

For a left-padded mask (`valid_start > 0`) this window contains `valid_start` fill rows (1e-10) followed by a shifted prefix of the real outputs, which line 274 then `index_copy_`s onto the mask positions — no error, wrong values. Affects the `logits_to_keep == 0` (embedding-mode) logits path and the hidden-states path. Note the runtime input side explicitly supports left padding (the base `_prepare_prefill_inputs` validates and compacts left- or right-padded masks, and generation *requires* left padding).

## 3. Chunk-loop variable `i` is shadowed → wrong `out_buffers` index

```python
186:  for i, step in enumerate(range(0, padded_input_length, ...)):   # chunk index
204:      for i in range(valid_len):                                   # shadows it
205:          cache_idx = cache_pos_chunk[0, i].item()
...
260:  super(...).forward(*forward_args, out=out_buffers[i])            # uses clobbered i
```

With `use_attention_mask=True` + `use_position_ids=True`, `out_buffers[i]` is indexed with `valid_len - 1` instead of the chunk index — IndexError on the first real prompt. Latent today (neither flag is forced for qwen3_vl) but breaks that configuration outright.

Minor: `modeling_qwen3_vl.py:742` reads `attention_mask[b_idx].sum(...)` without the `attention_mask is not None` guard that the call right below it has.

## Repro

Verified by driving the real `prefill_forward` with a stub `rebel` module and a fake runtime that writes position-coded values into the out buffers:

- right-padded mask + `output_hidden_states=True` → hidden states come back `(1, L, H)` instead of `(1, W, H)`; the model-level `copy_` then raises.
- left-padded mask → logits values wrong (fill rows scattered into the valid region).
- `use_attention_mask=True, use_position_ids=True`, 300-token prompt → `IndexError: list index out of range` at `out_buffers[i]`.

None of the existing tests catch these: the only multimodal `output_hidden_states` test is Gemma3 with a single unpadded prompt, and qwen3_vl has no `output_hidden_states` test at all.

rebel-thkim:
Fixes #658.

## What

Three defects in `RBLNQwen3VLRuntimeModel.prefill_forward` (all in the output-aggregation stage; same length-bookkeeping family as #649):

1. **Hidden states came back at valid length instead of mask width.** The runtime head-trimmed hidden states to `[:, :query_length]` while both model-level aggregations (`RBLNQwen3VLModel.forward` and the generation-path forward) `copy_` them into full-mask-width buffers — a shape-mismatch RuntimeError on any padded batch row.
2. **Left-padded masks produced silently wrong values.** `_prepare_prefill_outputs` writes valid rows at `[valid_start_index, valid_start_index + query_length)` of the out buffers, but the aggregation read the buffer head `[0, query_length)` — for left-padded masks that window is fill rows plus a shifted prefix, which `index_copy_` then scattered onto the mask positions.
3. **The chunk index `i` was shadowed** by the attention-mask update loop (`for i in range(valid_len)`), so `out=out_buffers[i]` indexed with `valid_len - 1` — IndexError under `use_attention_mask=True` + `use_position_ids=True`.

## Fix

A single `scatter_to_mask_width` helper: extract the valid window starting at `valid_start_index`, then `index_copy_` it onto the attention-mask positions of a mask-width tensor (fill 1e-10), applied to both the `logits_to_keep == 0` logits path and the hidden-states path. This preserves the existing right-padded behavior bit-for-bit (verified) while fixing left padding, and returns hidden states at the width the model level copies into — matching the logits path's own convention. The shadowed loop variable is renamed, and the model-level `query_length` computation gets the missing `attention_mask is not None` guard.

## Verification (no NPU needed)

Drove the real `prefill_forward` with a stub `rebel` and a fake runtime writing position-coded values into the out buffers:

- **After the fix**: all combinations pass — pad side (none/right/left) × valid lengths incl. exact chunk multiples × `output_hidden_states` on/off; logits and every hidden state come back `(1, mask_width, H)` with correct values at mask positions and 1e-10 elsewhere, and a model-style mask-width `copy_` succeeds. The `use_attention_mask + use_position_ids` path runs a 300-token multi-chunk prompt cleanly.
- **Before the fix** (same script): right-pad + hidden states returns `(1, 5, 8)` instead of `(1, 40, 8)`; left-pad returns wrong logits values; the shadowing path raises `IndexError: list index out of range`.
- Right-padded and unpadded logits outputs are identical before/after, so previously-working cases are unchanged.

## Test-suite hole

The only multimodal `output_hidden_states` test was Gemma3 with a single unpadded prompt; qwen3_vl had none. Added `TestQwen3VLForConditionalGeneration_OutputHiddenStates`: a batch-2, left-padded (two different-length prompts) generation test with `output_hidden_states=True`, reusing the existing `_test_output_hidden_states_generation` checks — its masked-rows-are-zero assertion also guards the left-pad scatter. The inherited `test_propagate_config` is skipped there to avoid a duplicate compile.


rebel-kblee (approved):
👍

rebel-seinpark (approved):

PR #662, opened 2026-08-03
model(qwen3.5): support output_hidden_states for the Qwen3.5 hybrid model

rebel-seinpark:
## What

Re-enables `output_hidden_states` for the Qwen3.5 hybrid model. It was previously
rejected at compile time with `NotImplementedError`; this wires the per-layer hidden
states through the whole stack (architecture graph → runtime → modeling forward).

## How

- **architecture** (`qwen3_5_architecture.py`): when `output_hidden_states` is set, the
  compiled graph returns `(logits, *new_states, *all_hidden_states)`. The per-layer
  activations (`n_layers + 1`) trail the linear-state (`conv`/`recurrent`) cache updates,
  so the runtime can slice them off the tail without knowing the `new_states` count.
- **runtime** (`qwen3_5_runtime_utils.py`): `_run` takes the last `n_layers + 1` outputs
  as hidden states. Prefill accumulates them per chunked window, trims to the prompt
  length, and scatters the valid tokens back to their attention-mask slots — mirroring the
  base runtime's `valid_start_index` placement (`_prepare_prefill_inputs` strips padding
  before the graph, so the graph only sees valid tokens). Decode returns the per-step
  hidden state.
- **modeling** (`modeling_qwen3_5.py`): the two VL forwards allocate a per-item
  `[batch, seq, hidden]` buffer and copy each item's hidden states in; the text
  `ForCausalLM` path reuses the base forward. Drops the `get_input_info` guard.

The non-OHS path is unchanged: with `output_hidden_states=False` the wrapper returns
`(logits, *new_states)` exactly as before, and the runtime returns `hidden_states=None`.

## Test

Added `TestQwen3_5ForConditionalGeneration_OutputHiddenStates` and ran the Qwen3.5 suite
(venv with rebuilt rebel-compiler, real 4-layer checkpoint):

```
23 passed, 1 skipped, 9 subtests passed in 148.75s
```

- `TestQwen3_5ForConditionalGeneration_OutputHiddenStates::test_generate` ✅ — validates
  shape, per-step count, and masked→zero / unmasked→nonzero placement.
- `TestQwen3_5ForConditionalGeneration` / `TestQwen3_5ForCausalLM` ✅ — no regression on
  the non-OHS path (incl. `test_save_load`). The single skip is `test_automap` (text model
  has `RBLN_AUTO_CLASS=None`, intentionally skipped).


src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
```diff
@@ -168,11 +177,38 @@ def prefill_forward(
...
+                full_len = attention_mask.shape[-1]
+                start = int(torch.nonzero(attention_mask.reshape(-1), as_tuple=False)[0].item())
+                placed = []
+                for vh in valid_hidden:
+                    buf = torch.zeros(1, full_len, vh.shape[-1], dtype=vh.dtype)
+                    buf[:, start : start + query_length, :] = vh
+                    placed.append(buf)
+                final_hidden_states = tuple(placed)
+            else:
+                final_hidden_states = tuple(valid_hidden)
```
rebel-kblee, line 205: can you check this one? @rebel-thkim

rebel-thkim: Checked this against the #658 bug family (the same aggregation pattern that bit qwen3_vl). **This implementation avoids both failure modes** — I traced it rather than just eyeballing:

1. **Left-padded masks**: the per-window `h[:, :valid_count, :]` head-trim is safe *here*, unlike the pre-fix qwen3_vl code. qwen3_5 windows run directly over the compacted input (`_prepare_prefill_inputs` strips padding before the graph), so each window's valid tokens really do start at row 0 — there is no `valid_start_index` offset like in the shared out-buffer views that caused #658. The scatter-back then places the concatenated `[1, query_length, H]` at `buf[:, start : start + query_length]` with `start` = first non-masked position, so left padding lands at the right slots.
2. **Model-level `copy_` width**: hidden states come back at full `attention_mask.shape[-1]` width (or `query_length` when there is no mask, which then equals `seq_len`), matching the `(batch, seq_len, H)` buffers both forwards copy into — no shape mismatch on padded batches (the #658 crash).

No `[:, :-padding]`-style slices in the new code either, so no #649-type empty-output hazard at chunk-multiple lengths (`valid_count >= 1` holds for every window since the padded length is `< query_length + chunk`).

Two things worth considering before merge:

- **Test coverage**: `TestQwen3_5ForConditionalGeneration_OutputHiddenStates` runs a single prompt at batch 1, so the two paths that silently broke in #658 — `start > 0` (left-padded row) and the padded-batch `copy_` — are never exercised on hardware. A batch-2, two different-length prompts, `padding_side="left"` variant (like `TestQwen3VLForConditionalGeneration_OutputHiddenStates` in #659) would cover both cheaply on top of this compile config.
- **Pre-existing, but surfaced by this PR**: `prefill_forward` still keeps only the *last window's* `logits`. For the bare `RBLNQwen3_5TextModel` (`logits_to_keep == 0`) that means `last_hidden_state` would be truncated to the final window on multi-chunk prompts, while the new `hidden_states[-1]` is the full sequence — the base contract expects these to be identical (`_test_output_hidden_states_generation` asserts exactly that). Harmless for the CausalLM / ConditionalGeneration paths (`logits_to_keep == 1`), but worth a follow-up if the bare text model is meant to be used directly.

Minor nit: `query_length = attention_mask[b_idx].sum(...)` in both model forwards still lacks an `attention_mask is not None` guard (same one-liner as fixed for qwen3_vl in #659).

rebel-thkim (approved):
👍

src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_runtime_utils.py
rebel-seinpark: **1. Test coverage** Done, with one caveat. `TestQwen3_5ForConditionalGeneration_OutputHiddenStates` now left-pads the prompt (`padding_side="left"`, `padding="max_length"`), so it runs with a non-zero start (`start=31`). The masked-zero assertion now exercises the scatter-back offset (`start > 0`) and the full-width `copy_` - the two #658 failure modes.

I kept it at **batch 1** rather than batch 2: the BC-inference job reuses a saved artifact keyed by the test class name and does not recompile, so a batch-size change makes a batch-2 input hit `block_tables[batch_idx=1]` on the reused batch-1 artifact (IndexError). Left padding at batch 1 covers the `start > 0` path without that mismatch. The multi-row `copy_` (`b_idx > 0`) is therefore not exercised here - happy to add a separate batch-2 class if we can give it its own BC artifact.

**2. last_hidden_state truncation**  Agreed it is pre-existing and only affects the bare `RBLNQwen3_5TextModel` (`logits_to_keep == 0`); CausalLM / ConditionalGeneration use `logits_to_keep == 1`. Tracking it as a follow-up.

PR #663, opened 2026-08-04
performance: unify host rotary cos/sin into deterministic lookup tables

rebel-thkim:
## Summary

Stacked PR 2/2 on #672 (np_cos/np_sin conversion). That PR makes host rotary cos/sin deterministic; this one unifies **how** they are produced: cos/sin tables built once at load (via `np_cos`/`np_sin`, so still bit-reproducible) and gathered per call — the approach #655/#653 introduced per-model, now shared across the qwen family.

Lookup is bit-identical to the dynamic numpy path (`cos(table)[ids] == cos(table[ids])`, elementwise) and faster at every size/thread count on the prefill path (2–4× vs torch, 4–10× vs np dynamic).

## Change

- **LLM host rotary → `QwenMRopeLookupTable`** (`modeling_rope_utils.py`): the `_get_position_embeddings` path of qwen2/2.5/3-vl(+moe)/3.5 computed cos/sin on host via HF rotary modules at prefill & every decode step. `build_qwen_mrope_lookup` wraps the module with a table sized by the compiled `max_seq_len` (1 KB/row at head_dim=128, e.g. 114,688 → 112 MB). Supports both mrope layouts — standard `[3,bs,seq,dim]` and interleaved (per-frequency axis selection baked into the gather index, mirroring `apply_interleaved_mrope`). Out-of-range positions (e.g. mrope temporal ids) fall back to the bit-identical dynamic numpy path. Any rope type with static `inv_freq` qualifies (default/linear/yarn/llama3); `dynamic`/`longrope` keep the original module with a warning.
- **Vision rotary → lookup tables**: the qwen2/2.5/3-vl(+moe)/exaone4_5 vision towers gather from tables built once at load (same approach as qwen3_5 in #655), with the shared `qwen_vit_rot_pos_ids` replacing three duplicated (h, w) pos-id builders.
- Helper naming is scoped honestly: `np_cos`/`np_sin` stay generic (any model can use them directly — gemma4 does); the mrope/ViT helpers carry the qwen prefix.

## Verification (Qwen2.5-VL-7B, RBLN-CA25 ×8, bf16)

- **Bit-identity**: all 5 rotary classes (`Qwen2VL`, `Qwen2_5_VL`, `Qwen3VLText`, `Qwen3VLMoeText`, `Qwen3_5Text`) — lookup vs HF module ≤ 1 ulp across prefill/decode/2-D shapes; table gather vs dynamic fallback **bit-equal**; vision old-path vs table-path bit-equal including window reorder; E2E `generated_ids` **unchanged** on NPU.
- **Function-level host time** (cold, p50 of 3 runs; np → lookup):

| site | case | 48 threads |
|---|---|---|
| LLM prefill | 27,305 tokens | 85 → **9** ms |
| LLM prefill | 114,367 tokens (video) | 336 → **82** ms |
| LLM decode | per token | ~101 → ~90 µs |
| vision | 457,368 patches | 161 → 62 ms |

- **E2E**: host cos/sin share drops further (27k prefill 26.5 → 14.7 ms, video 18.8 → 7.3 ms); E2E totals unchanged within noise (host share < 1% everywhere). 256k-table sanity verified (262,144 rows, 256 MB, ≤ 1 ulp).


src/optimum/rbln/transformers/models/exaone4_5/modeling_exaone4_5.py
```diff
@@ -245,13 +249,13 @@ def forward(self, hidden_states: torch.Tensor, grid_thw: torch.Tensor) -> torch.
         hidden_states = hidden_states.reshape(seq_len // self.spatial_merge_unit, self.spatial_merge_unit, -1)
         hidden_states = hidden_states[window_index, :, :]
         hidden_states = hidden_states.reshape(seq_len, -1)
-        rotary_pos_emb = rotary_pos_emb.reshape(seq_len // self.spatial_merge_unit, self.spatial_merge_unit, -1)
-        rotary_pos_emb = rotary_pos_emb[window_index, :, :]
-        rotary_pos_emb = rotary_pos_emb.reshape(seq_len, -1)
-        emb = torch.cat((rotary_pos_emb, rotary_pos_emb), dim=-1)
+        pos_ids = pos_ids.reshape(seq_len // self.spatial_merge_unit, self.spatial_merge_unit, 2)
+        pos_ids = pos_ids[window_index, :, :].reshape(seq_len, 2)
```
rebel-seinpark: nit) reshape(..., 2) here is equivalent to -1 (since pos_ids is [*, 2] and `seq_len % spatial_merge_unit == 0`). Either use -1 to match the `hidden_states` line right above, or keep 2 with a short `# (h, w) axes` comment. Not a correctness issue.

rebel-thkim: Changed to `-1` to match the hidden_states lines — applied to the same pattern in qwen2_5_vl as well. (e7f58c0b)

PR #666, opened 2026-08-11
refactor: split cache metadata into a polymorphic CacheMeta hierarchy

rebel-seinpark:
## Summary

Refactors the decoder-only cache metadata from a single string-typed `KVCacheMeta` dataclass into a
small **polymorphic `CacheMeta` hierarchy**, moved to `transformers/cache_utils.py` (mirroring HF
Transformers' `cache_utils.py` location).

```
CacheMeta                         # base (abstract): name, layer_index, shape, dtype; serialization
├─ KVCacheMeta                    # paged KV (abstract): num_blocks / block_size
│   ├─ FullAttentionKVCacheMeta   # resizable when is_auto
│   └─ SlidingWindowKVCacheMeta   # fixed-size window
└─ LinearAttentionCacheMeta       # conv/recurrent state; raw model-computed shape, never resized
```

## Motivation

The previous single `KVCacheMeta` dataclass crammed three different cache kinds together:

- `can_resize` was a **magic-string** property: `is_auto and layer_type == "full_attention"`.
- `shape` was documented as 4D `[num_blocks, num_heads, block_size, head_dim]`, but linear-attention
  (Qwen3.5 GatedDeltaNet) stuffed raw `conv_state` / `recurrent_state` shapes into the same field via a
  `shape=` early-return hack in `KVCacheMeta.make()`.
- Linear-attention state was only registered as a `KVCacheMeta` (with `layer_type="linear_attention"`)
  as a **stopgap** so the block-estimation code's `can_resize[key]` lookup wouldn't `KeyError` on the
  linear DRAM tensors.

## What changed (structure)

- Each cache kind is now its own type; `can_resize` / `compile_shape` are **per-subclass properties**
  (no magic string), and `layer_type` is a class-level tag.
- Construction goes through per-subclass `from_config()` factories. The old `KVCacheMeta.make()`
  if-chain dispatcher is **removed** — each builder picks the concrete subclass by the layer kind it
  already knows (decoder-only base & gemma4: sliding vs full; qwen3.5: linear vs full).
- `CacheMeta` / `KVCacheMeta` are abstract (`from_config` is `@abstractmethod`; ABCMeta comes from the
  serialization Protocol base) — only the concrete leaves are instantiable.
- Linear state is now a first-class `LinearAttentionCacheMeta` instead of a disguised `KVCacheMeta`, so
  the block-estimation `KeyError` is avoided **by design** rather than by the stopgap registration.
- `RBLNDecoderOnlyModelConfig.kvcache_metas` is renamed to **`cache_metas`** (the list now holds non-KV
  linear metas too, so `kvcache_metas` was a misnomer). A `@deprecate_kwarg(old_name="kvcache_metas",
  new_name="cache_metas", version="0.12.0")` remaps the old key.

## Behavior

Runtime and block-estimation behavior is **unchanged**; the one observable difference is the serialized
config key rename, handled with a deprecation remap.

| Aspect | Before | After |
| --- | --- | --- |
| `can_resize` | `is_auto and layer_type == "full_attention"` | subclass property (Full = `is_auto`; sliding/linear = `False`) — **identical values** |
| `compile_shape` | `[1, h, block, d]` if resizable else `shape` | Full flattens blocks→1 when `is_auto`; others return `shape` — **identical** |
| `num_block` estimation / DRAM sizing | `_kvcache_bytes_per_chiplet` scales resizable tensors; `multiply_kv_cache_num_blocks` grows `can_resize` buffers | same inputs (`meta.name` / `meta.can_resize` / `compile_shape`) — **identical result** |
| Config field / serialized key | `kvcache_metas` | **`cache_metas`** (old key remapped on load via `@deprecate_kwarg`) |
| Per-meta serialized schema | `{name, layer_index, shape, layer_type, is_auto, dtype}` | same keys emitted (`is_auto` defaults to `False`) — **identical** |

Why the estimation is unaffected: `modeling_attention_utils` reads only `meta.name` and
`meta.can_resize` (plus `compile_shape` indirectly, via what gets compiled). All three are byte-identical
to the old formula across every cache kind (verified: full-auto / full-manual / sliding / linear-conv /
linear-recur). Avoiding the `can_resize[key]` `KeyError` for linear tensors is now structural (linear is
a registered first-class meta) instead of the previous stopgap.

## Backward compatibility

- **Config key rename** `kvcache_metas` → `cache_metas` is remapped on load by `@deprecate_kwarg`, so
  existing `rbln_config.json` artifacts (which serialize `kvcache_metas`) load unchanged. Metas are
  compile-time only and inert on load anyway.
- The per-meta serialized schema is unchanged; `layer_type` is retained purely as a serialized tag for
  readability/compat and no longer drives any logic.

## Testing

- Unit parity check: new `can_resize` / `compile_shape` / serialized schema match the old formula for
  all five cache kinds; `@deprecate_kwarg` remaps `kvcache_metas=` → `cache_metas=`.
- Compile + generate tests pass: `TestQwen3_5ForConditionalGeneration` (full + linear),
  `TestGemma3ForCausalLM` (sliding + full) — covers the decoder-only base builder and the qwen3.5 builder.
- `ruff check` / `ruff format` clean.

### DIFF
--- a/src/optimum/rbln/transformers/cache_utils.py
+++ b/src/optimum/rbln/transformers/cache_utils.py
@@ -1,0 +1,187 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Metadata classes describing the on-device caches of a decoder-only model.
+
+Analogous to Transformers' ``cache_utils`` (which keeps ``DynamicCache`` / sliding-window /
+linear-attention layers as distinct types), this module holds a small polymorphic hierarchy:
+
+    CacheMeta                     # base: name, layer_index, shape, dtype; serialization
+    ├─ KVCacheMeta                # paged KV: [num_blocks, num_heads, block_size, head_dim]
+    │   ├─ FullAttentionKVCacheMeta      # resizable when is_auto (grown after compile)
+    │   └─ SlidingWindowKVCacheMeta      # fixed-size window
+    └─ LinearAttentionCacheMeta   # conv/recurrent state; raw model-computed shape, never resized
+
+Each subclass owns its ``can_resize`` / ``compile_shape`` and a class-level ``layer_type`` tag, so
+the previous single-dataclass-with-a-``layer_type``-string design (and its magic-string ``can_resize``)
+is gone. Instances are built through the ``make()`` dispatcher / the subclass ``from_config()``
+factories, which compute the derived ``shape`` and construct the meta.
+"""
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, ClassVar
+
+from ..configuration_utils import RBLNSerializableConfigProtocol
+
+
+if TYPE_CHECKING:
+    from .models.decoderonly.configuration_decoderonly import RBLNDecoderOnlyModelForCausalLMConfig
+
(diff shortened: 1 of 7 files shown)


rebel-jongho (approved):
LGTM

PR #668, opened 2026-08-13
model: qwen3.5 aggregate full-sequence logits for the bare text model

### LINKED ISSUE #667: Qwen3.5 bare TextModel returns only the last prefill window as last_hidden_state
Found while reviewing #662.

`RBLNQwen3_5RuntimeModel.prefill_forward` keeps only the last window's graph output as `logits` (correct for `logits_to_keep == 1`, where every window overwrites the single next-token row). But for the bare `RBLNQwen3_5TextModel` (`logits_to_keep == 0`, no LM head), the graph emits full-chunk hidden states as "logits", so the returned `last_hidden_state` is just the final window (chunk width, including right-pad rows) instead of the full sequence:

- prompts longer than `prefill_chunk_size` → all earlier windows are dropped;
- even single-window prompts → width is the padded chunk size, not `query_length`.

Once #662 lands, this also becomes an observable contract violation: `output.hidden_states[-1]` (aggregated across windows) will be the full sequence while `output.last_hidden_state` is not, and `_test_output_hidden_states_generation` asserts they match.

Mitigating factors: `RBLNQwen3_5TextModel` is currently dormant — exported only from the `qwen3_5` subpackage (not `optimum.rbln` top level), no tests, not used by vllm-rbln. So nothing reachable through the public API hits this today.

Suggested fix (after #662 merges, since it adds the needed machinery): when `logits_to_keep == 0`, aggregate per-window logits the same way #662 aggregates hidden states — collect `[:, :valid_count]` per window, concat, and scatter back to the attention-mask width — instead of keeping only the last window.

rebel-thkim:
Fixes #667. **Stacked on #662** — the fix reuses the per-window aggregation #662 introduces for hidden states, so this PR targets the `qwen3-5-output-hidden-states` branch and should merge after (or into) it. GitHub will retarget it to `dev` automatically if #662 merges first and its branch is deleted.

## What

For `logits_to_keep == 0` (the bare `RBLNQwen3_5TextModel`, no LM head) the prefill graph emits full-chunk hidden states as "logits", but `prefill_forward` kept only the last window's output. As a result `last_hidden_state` came back at padded chunk width for every prompt (e.g. `(1, 128, H)` for a 5-token prompt) and, for multi-chunk prompts, contained only the final window. Once #662 lands this also violates the base contract that `last_hidden_state == hidden_states[-1]`.

## Fix

When `logits_to_keep == 0`, collect each window's valid rows (`[:, :valid_count]`), concat along the sequence, and scatter back to the attention-mask slots — the exact aggregation #662 added for hidden states, now shared as a `place_at_mask_slots` helper used by both paths. `logits_to_keep == 1` (generation) is untouched: still the single next-token row from the last window.

## Verification (no NPU needed)

Drove the real `prefill_forward` with a stub `rebel` and a fake runtime returning position-coded outputs through the named-arg protocol:

- **After the fix**: `logits_to_keep=0` returns the full sequence at mask width for every combination — pad side (none/right/left) × single/multi-window lengths incl. exact chunk multiples × `output_hidden_states` on/off — with correct per-position values, zeros at masked rows, and `last_hidden_state == hidden_states[-1]` holding. `logits_to_keep=1` returns the identical last-token row as before.
- **Before the fix** (same script on the #662 head): even the simplest case fails — a 5-token prompt returns logits of shape `(1, 128, 8)`.

## Note

`RBLNQwen3_5TextModel` is currently dormant (exported only from the `qwen3_5` subpackage, no tests, unused by vllm-rbln), so there is no CI test exercising this path yet; adding a top-level export + a `TestLLMWithoutLMHead`-based test class would be a reasonable follow-up once the class is meant to be public.


rebel-seinpark:
@rebel-thkim Please resolve conflicts.

rebel-seinpark (approved):

PR #670, opened 2026-08-13
other: remove obsolete vllm model-generation dispatch

rebel-seinpark:
## What

Remove the `trigger-vllm-model-compilation` job from `deploy-on-tag.yaml`.

## Why

The job dispatched `rebel_dispatch_model_generation_for_vllm.yaml` in `rebellions-sw/rebel_compiler`, but that workflow no longer exposes a `workflow_dispatch` trigger. So the step fails on **every tag deploy**:

```
RequestError [HttpError]: Workflow does not have 'workflow_dispatch' trigger
POST .../rebel_compiler/actions/workflows/rebel_dispatch_model_generation_for_vllm.yaml/dispatches
```

(e.g. run 31694450151)

The dispatch target is gone, so the trigger is dead weight. No other job `needs` it, and the remaining jobs (`check-tag`, `deploy-on-tag`, `trigger-vllm-rbln-pkg-update`, `bc-compile`) are unaffected.

### DIFF
--- a/.github/workflows/deploy-on-tag.yaml
+++ b/.github/workflows/deploy-on-tag.yaml
@@ -59,28 +59,6 @@ jobs:
             })
             console.log(result)
 
-  trigger-vllm-model-compilation:
-    name: trigger vllm model compile and generation ci
-    needs: [deploy-on-tag]
-    runs-on: rbln-sw-k8s-4c-32gb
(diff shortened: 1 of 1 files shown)


rebel-junamsong (approved):
LGTM

PR #653, opened 2026-07-30
model: gemma4 bug fix

rebel-dkhong:
# Pull Request Description

> **⚠️ Important: Branch Target**
> - **New features, enhancements, and non-critical fixes**: Merge to `dev` branch
> - **Critical hotfixes only**: Merge to `main` branch (must also merge to `dev`)
> 
> Please ensure you've selected the correct base branch before submitting!

## Type of Change

- [ ] Release (dev → main merge for production release)
- [ ] New Model Support
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe):

## Changes Overview

1. To prevent unexpected mismatch while computing rope, calculate them before inference and reuse.
2. Set input dtype of vision_tower & embed vision according with rbln_config.dtype

## Related Issues

----
# Conventional commit
```
type(optional scope): description
```
----
# Type candidate
  - Model Updates
    - `model`: Adding New models or Bugfix for existing models
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
    - `performance`: Optimizing some models or this library itself
      - ex) Loading RBLNModel faster
      - ex) Optimizing Memory Usage of DecoderOnlyModel
  - Code Refactor
    - `refactor`: Re-arrange class architecture, or more.
      - ex) Refactor Seq2Seq
  - Documentation
    - `doc`: Update docstring only
  - Library Dependencies
    - `dependency`: Update requirements, something like that.
  - Release
    - `release`: Merging dev to main for production release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

PR #672, opened 2026-08-19
performance: compute host-side rotary cos/sin deterministically via numpy

rebel-thkim:
## Summary

torch's CPU `cos()`/`sin()` kernels are not bit-reproducible: results can differ in the last bits depending on the OMP/MKL thread configuration and tensor size (environment-dependent). Host-computed rotary cos/sin fed to compiled graphs as inputs therefore breaks run-to-run reproducibility. #655 and #653 worked around this per-model; this PR fixes the root cause generically.

Stacked PR 1/2 — #663 (lookup-table unification) builds on top of this branch.

## Change

- New helpers `np_cos(x)` / `np_sin(x)` in `transformers/modeling_rope_utils.py` — tensor → numpy → `np.cos`/`np.sin` → tensor. numpy's cos/sin are reproducible across thread counts, and the tensor↔numpy conversion is **zero-copy** (same memory, ~1.6 µs fixed, size-independent).
- Applied at every host-side cos/sin site: **qwen2_vl / qwen2_5_vl / qwen3_vl / exaone4_5** vision forwards, **qwen3_5** vision tables (#655 follow-up), **gemma4** vision tables (#653 follow-up, verified 1-ulp identical to the HF-module path), **decoderonly `RotaryEmbedding`** (compile-time constants → reproducible artifacts).

## Verification — experiments on Qwen2.5-VL-7B (RBLN-CA25 ×8, zoo compile config, bf16)

Full design & scripts: `cos_deterministic.md` / `cos_deterministic_report.md` (experiment report, kept out of this PR).

**1. Equivalence (np vs torch cos)**
- numpy determinism: **PASS — 30/30 identical sha256** across threads {1,2,4,8,48} × repeats × full/row-wise slicing. (torch non-determinism did not reproduce on the test machine — it is environment-dependent.)
- Accuracy vs float64 over the real rope domain (0–114,688 rad): torch max 3.6e-08, numpy max 6.9e-08 — same class (≈1 ulp). Mutual diff: max **1 ulp**.

**2. Function-level host time** (real position_ids from the processor; cold cache, p50 of 3 runs; **A=torch / B=np (this PR)**)

| site | case | 1 thread: A / B | 48 threads: A / B |
|---|---|---|---|
| vision (per image) | 5,120 patches | 0.95 / 0.40 ms | 0.14 / 0.41 ms |
| vision (per video) | 457,368 patches | 197 / **154** ms | **43** / 161 ms |
| decode (per token) | `[3,1,1]` | ~134 / ~101 µs (cold) | same |

np dynamic loses to torch only with **many threads × large tensors** — and that parallelism is exactly the source of the non-determinism (single-threaded, np beats torch everywhere). The large-tensor gap is eliminated by the lookup tables in the stacked PR #663.

**3. E2E on NPU** (same compiled artifact, host code only swapped; greedy, 3 runs p50)

| UC (prefill len) | E2E total A / B (s) | host cos/sin share A / B (%) |
|---|---|---|
| image 1k patches (281) | 1.14 / 1.14 | 0.65 / 0.65 |
| image 5k patches (1,305) | 1.39 / 1.41 | 0.57 / 0.88 |
| image + long text (27,317) | 6.88 / 6.85 | 0.52 / 0.39 |
| video 102k patches (25,513) | 119.1 / 119.1 | 0.011 / 0.016 |

- **host cos/sin share is <1% in every case; E2E totals are identical within noise → no user-visible regression.**
- Output equality (greedy): 3 of 4 UCs produce **bit-identical `generated_ids`**. The long-text UC flips one token: ~0.0003% of elements keep a 1-ulp diff even after bf16 cast (22/7M measured), which can flip a near-tie argmax. This is a baseline *fix*, not a regression — the np value is the one reproducible across environments. Golden-output tests should use correlation, not exact match.


rebel-seinpark (approved):
LGTM

PR #674, opened 2026-08-19
model(decoderonly): load embed_tokens in the compile dtype

rebel-eunji:
## Type of Change

## Changes Overview
- Build `embed_tokens` in `rbln_config.dtype` at construction, inside every `_create_embedding_layer`: the plain `nn.Embedding` sites take a `dtype` kwarg, and `load_state_dict` casts the fp32 checkpoint into the module dtype. The Gemma scaled-embedding classes (`Gemma3/Gemma4TextScaledWordEmbedding`) do not forward a `dtype` kwarg, so those cast the module after construction.
- Move gemma4's existing load-time casts into its `_create_embedding_layer` / `_create_per_layer_embedding_layer` accordingly.
- Drop the call-site casts that are now no-ops: `self.embed_tokens(input_ids).to(self.rbln_config.dtype)` in qwen2_vl, qwen2_5_vl, qwen3_vl, qwen3_5 (prefill/decode each) and exaone4_5 (prefill). The cast in Exaone-4.5's `forward` stays because it also covers caller-passed `inputs_embeds`.

## Motivation and Context
`_create_embedding_layer` built an fp32 `nn.Embedding` and `load_state_dict` keeps the module dtype, so a model compiled with bf16/fp16 still served its embedding lookups in fp32.

Consequences:
- Every consumer had to patch over it at the call site with `.to(rbln_config.dtype)`; the same patchwork leaked into vllm-rbln (its `embed_tokens(...).to(...)` casts become removable once this lands).
- Paths without the patch fed fp32 embeddings to a runtime compiled for bf16/fp16 — e.g. `decoderonly_runtime_utils.py`'s inputs_embeds preparation returns `self.embed_tokens(input_ids)` uncast.

Applying the dtype inside `_create_embedding_layer` (rather than after load in `__post_init__`) makes the invariant hold wherever the layer is constructed, including subclasses with their own load paths. For fp32-compiled models (including all `_supports_non_fp32 = False` classes) it is a no-op, so behavior is unchanged there.

`_get_position_embeddings` casts cos/sin to `rbln_config.dtype` itself, so removing the call-site casts does not affect the position-embedding dtype.

## Related Issues
Counterpart of the vLLM/optimum dtype sync work in vllm-rbln (fix/model-dtype-redo).

### DIFF
--- a/src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
+++ b/src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
@@ -82,10 +82,6 @@ class RBLNDecoderOnlyModel(RBLNModel, RBLNDecoderOnlyFlashAttentionMixin):
             artifacts = torch.load(self.model_save_dir / self.subfolder / "torch_artifacts.pth", weights_only=False)
             self.embed_tokens = self._create_embedding_layer()
             self.embed_tokens.load_state_dict(artifacts["embed_tokens"])
-            # _create_embedding_layer builds an fp32 module and load_state_dict
-            # keeps that dtype, so cast to the compile dtype: the lookup output
-            # feeds the runtime's inputs_embeds, which expects rbln_config.dtype.
-            self.embed_tokens.to(self.rbln_config.dtype)
         else:
             self.embed_tokens = None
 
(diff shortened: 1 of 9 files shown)


rebel-kblee:
@rebel-thkim @rebel-seinpark @rebel-dkhong Can you think of any issues that might arise if we change `embed_tokens` as described in this PR? It seems it should be fine as long as the func issue is resolved—what do you think?

rebel-thkim:
How about fixing it within the `_create_embedding_layer` function itself?

src/optimum/rbln/transformers/models/decoderonly/modeling_decoderonly.py
```diff
@@ -198,11 +198,15 @@ def save_torch_artifacts(
             torch.save(save_dict, save_dir_path / subfolder / "torch_artifacts.pth")
 
     def _create_embedding_layer(self):
+        # Created in the compile dtype: load_state_dict casts the fp32
+        # checkpoint into it, and the lookup output feeds the runtime's
+        # inputs_embeds, which expects rbln_config.dtype.
```
rebel-eunji:
```suggestion
```

rebel-eunji:
@rebel-thkim Great Idea. I applied your comment.

rebel-kblee (approved):
it looks good to me! 👍

rebel-seinpark (approved):
LGTM!

PR #676, opened 2026-08-19
model(llava): size the projector buffer by images, not images x channels

rebel-junamsong:
## Symptom

Serving `llava-hf/llava-1.5-7b-hf` with one image fails at the first request:

```
ValueError: Multimodal placeholder/embedding count mismatch:
576 placeholder positions but 1728 embed tokens
```

576 = (336 / 14)² is the correct count for one LLaVA-1.5 image, and the model's own config agrees (`image_seq_length: 576`, `vision_config.image_size: 336`, `patch_size: 14`). 1728 is exactly **3×** that.

## Cause

`RBLNLlavaForConditionalGeneration.get_image_features` pre-allocates the projector output as

```python
projector_out_size = [
    pixel_values.shape[0] * pixel_values.shape[1],
    (image_size // patch_size) ** 2,
    text_hidden_size,
]
```

For LLaVA-1.5 `pixel_values` is `(num_images, 3, H, W)`, so **dim 1 is the channel count**. With one image the leading dimension becomes `1 × 3 = 3`, the buffer is three times too large, and the call returns three images' worth of features — which is where the 3× comes from.

Two things confirm the reading:

- **`vision_out_size`, a few lines above in the same function, uses `pixel_values.shape[0]` alone.** The two buffers disagreed about what dim 1 means.
- **The expression is the LLaVA-NeXT shape.** There `pixel_values` is `(batch, num_patches, 3, H, W)` — `modeling_llava_next.py` branches explicitly on `pixel_values.dim() == 5` — so `shape[0] * shape[1]` is right, and LLaVA-NeXT serves correctly. It looks carried over into the 4-D path.

## Fix

Use `pixel_values.shape[0]`, matching `vision_out_size` in the same function. One line, plus a comment recording why dim 1 must not be multiplied in here.

## Verification

Reproduced end to end by serving the documented recipe command on RBLN-CA22 (16 devices, KMD 3.2.2) with `vllm 0.22.0+cpu` + `vllm-rbln 0.11.1`:

```
VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK=4 \
vllm serve llava-hf/llava-1.5-7b-hf \
    --served-model-name llava-1.5-7b-hf \
    --max-model-len 4096 --block-size 4096 --max-num-seqs 1
```

The mismatch is deterministic on the first request. `llava-next-mistral-7b`, `qwen2-vl-7b`, `idefics3-8b` and `qwen2.5-vl-7b` all served images fine in the same sweep, which is what isolated this to the 4-D LLaVA path rather than multimodal serving in general.

**No test:** the function needs a compiled RBLN vision tower, so there is no unit seam for the shape arithmetic in the current test layout (`tests/test_transformers.py` instantiates and compiles). Happy to add one if you would like a mocked vision tower for this.

## Provenance

Found while running all 53 documented vLLM recipes from `rbln_sdk_doc` on hardware, verbatim as a reader would copy them.

### DIFF
--- a/src/optimum/rbln/transformers/models/llava/modeling_llava.py
+++ b/src/optimum/rbln/transformers/models/llava/modeling_llava.py
@@ -386,7 +386,13 @@ class RBLNLlavaForConditionalGeneration(RBLNModel, RBLNDecoderOnlyGenerationMixi
             image_features = projected_features[:, :num_real_patches, :]
         else:
             projector_out_size = [
-                pixel_values.shape[0] * pixel_values.shape[1],
+                # Images, not images x channels: this pixel_values is
+                # (num_images, 3, H, W), so dim 1 is the channel count. Multiplying
+                # by it (the LLaVA-NeXT shape, where dim 1 is num_patches) sized the
+                # buffer 3x too large, and get_image_features returned three images'
+                # worth of features for one image. vision_out_size above already
+                # uses shape[0] alone.
+                pixel_values.shape[0],
                 (self.config.vision_config.image_size // self.config.vision_config.patch_size) ** 2,
                 self.config.text_config.hidden_size,
             ]
--- a/src/optimum/rbln/transformers/models/llava/modeling_llava.py
+++ b/src/optimum/rbln/transformers/models/llava/modeling_llava.py
@@ -386,12 +386,7 @@ class RBLNLlavaForConditionalGeneration(RBLNModel, RBLNDecoderOnlyGenerationMixi
             image_features = projected_features[:, :num_real_patches, :]
         else:
             projector_out_size = [
-                # Images, not images x channels: this pixel_values is
-                # (num_images, 3, H, W), so dim 1 is the channel count. Multiplying
-                # by it (the LLaVA-NeXT shape, where dim 1 is num_patches) sized the
-                # buffer 3x too large, and get_image_features returned three images'
-                # worth of features for one image. vision_out_size above already
-                # uses shape[0] alone.
+                # pixel_values is (num_images, channels, height, width).
                 pixel_values.shape[0],
                 (self.config.vision_config.image_size // self.config.vision_config.patch_size) ** 2,
                 self.config.text_config.hidden_size,


rebel-junamsong:
Reproduced on the release-candidate stack as well, with the same 3x: `vllm 0.24.0+cpu` + `vllm-rbln 0.11.2a9.dev11+gd6371b63e` + `optimum-rbln 0.11.2a1` gives

```
ValueError: Multimodal placeholder/embedding count mismatch:
576 placeholder positions but 1728 embed tokens
```

identical to `0.22.0+cpu` / `0.11.1` / `0.11.1`. So this is not version-specific — `pixel_values.shape[0] * pixel_values.shape[1]` is present in both, and a version bump does not clear it.

For contrast, in the same retest sweep every failure that was environmental (leaked device memory, a full disk during serve-time compilation) turned into a pass on this stack, and the two that reproduced unchanged were this one and a separate rebel-compiler MoE sharding error. That separation is what isolates this to the code path rather than the environment.

rebel-junamsong:
Why the rbln-model-zoo example does not show this, since it is a fair question about whether the bug is real:

Both paths take the same branch. `llava-1.5-7b/compile.py` sets only `vision_tower.output_hidden_states` and `language_model.{num_devices, use_inputs_embeds}` — no `max_image_size` — so `get_image_features` falls into the same `else` that sizes the projector buffer by `shape[0] * shape[1]`. The 3x buffer is allocated in the model-zoo run too.

The difference is entirely in who consumes the return value.

- This class's own `forward` merges it at `modeling_llava.py:438-440` with `inputs_embeds.masked_scatter(special_image_mask, image_features)`. `masked_scatter` reads only as many elements from the source as the mask has `True` positions, in row-major order — that is exactly the first image's 576x4096, which is the correct data. The two extra images' worth of rows are never read. Verified: scattering the correct `[1, 576, H]` and an oversized `[3, 576, H]` whose tail is NaN produce a bit-identical result, no NaN reaches the embeddings.
- vllm-rbln consumes it as a batch: `model_executor/models/optimum/llava.py:147` returns `list(image_features)`, which iterates dim 0 and yields three per-image embeddings for one image. vLLM then counts them against the placeholders and raises `576 placeholder positions but 1728 embed tokens`.

So the model-zoo path is correct by accident — the wrong size is masked away rather than absent. What is left even there is a 3x over-allocation per image and uninitialized rows handed to any caller that reads the batch dim. `vision_out_size` a few lines above already uses `shape[0]` alone, and the multiplying form is the LLaVA-NeXT shape where dim 1 really is `num_patches`, which is why LLaVA-NeXT serves fine.

rebel-thkim (approved):
👍

PR #678, opened 2026-08-20
model(gemma4): strip image padding by the position ids, not the returned mask

rebel-junamsong:
## Symptom

Serving gemma-4 with an image on the **dummy device** (`rbln_config.device = -1`) fails at the caller's per-image split:

```
RuntimeError: split_with_sizes expects split_sizes to sum exactly to 280
              (input tensor's size at dimension 0), but got split_sizes=[266]
```

This is how the deploy vLLM suite runs — a suite step gets one NPU while these recipes ask for sixteen, so it validates at the API level on the dummy device. `gemma-4-26b-a4b` and `gemma-4-31b` have been red there since they were added, while **the same documented commands serve and answer on 16× RBLN-CA22** (both models, on two stacks).

## Measured, not inferred

I instrumented the strip rather than reasoning about it — two earlier explanations of mine were wrong:

```
[probe] mask shape=(1, 280) dtype=torch.bool true=280  pooler_out=(1, 280, 2816)
[probe] after mask -> (280, 2816)
[probe] image_embeds=(280, 2816) valid_lens=[266] k2=9 valid_patches=[2394]
```

- 2394 valid patches / 3² = **266** real soft tokens
- 2520 max patches / 3² = **280**, the compiled bucket
- the tower's `pooler_mask` marks **280 of 280** valid, so the strip removes nothing and 14 padding slots survive

What it is **not** (each ruled out by experiment, not by reading):

| hypothesis | test | result |
|---|---|---|
| deploy's layer reduction | run with reduction full / text-only / **absent** | fails identically all three |
| the mask buffer is left unwritten | seed it with the correct mask before the call | unchanged — the runtime overwrites it |

## Fix

Derive the mask from `pixel_position_ids`, where `(-1, -1)` already marks padding. That is the same arithmetic vllm-rbln uses to split the returned features, so producer and consumer agree **by construction** rather than by assumption. On device the two masks match, so behaviour there is unchanged; a disagreement is logged rather than swallowed, so a genuine tower bug stays visible.

## Verification

- **End to end:** the documented `gemma-4-26b-a4b` serve command on the dummy device went from failing that split to **serving and answering**. (Output text from a dummy device is meaningless by construction — this is API-level validation.) The warning fired once, naming the measured numbers.
- **On device:** the same commands already pass on 16× RBLN-CA22 for both models on two stacks, so this touches only the path that was broken.
- **Unit:** `tests/test_gemma4_pooler_mask.py` — the CI case (2394/2520 → 266/280), two images, no padding, all padding, and that valid tokens sit at the front, which the caller's split depends on. 5 passing.

`ruff check` and `ruff format` clean.

## Related

[rbln_deploy#697](https://github.com/rebellions-sw/rbln_deploy/pull/697) stops the suite reporting this as a broken recipe. If this lands, that tolerance becomes unnecessary for gemma-4 — happy to close it in favour of this.

### DIFF
--- a/src/optimum/rbln/transformers/models/gemma4/modeling_gemma4.py
+++ b/src/optimum/rbln/transformers/models/gemma4/modeling_gemma4.py
@@ -595,12 +595,8 @@ def pooler_mask_from_position_ids(
     """Which soft-token slots of the pooler output hold a real patch block.
 
     `pixel_position_ids` is (batch, max_patches, 2) with (-1, -1) marking padding,
-    and the pooler averages each `pooling_kernel_size**2` block into one soft
-    token, so row i holds `valid_patches(i) // k**2` real tokens at the front.
-
-    This is the same arithmetic vllm-rbln uses to split the returned features per
-    image, so deriving the mask here keeps the producer and the consumer in
-    agreement by construction rather than by assumption.
+    and the pooler averages each `pooling_kernel_size**2` block into one soft token,
+    so row i holds `valid_patches(i) // k**2` real tokens at the front.
     """
     valid_patches = (~(pixel_position_ids == -1).all(dim=-1)).sum(dim=-1)
     valid_soft_tokens = valid_patches // (pooling_kernel_size * pooling_kernel_size)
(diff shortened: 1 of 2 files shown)


rebel-junamsong:
Closing in favour of a deploy-side correction: rebellions-sw/rbln_deploy#699.

The reasoning that changed my mind is that `device: -1` is a developer convenience, not a path a customer takes. The documented gemma-4 command serves and answers on 16x RBLN-CA22 — the geometry is 2394 of 2520 patches with 3x3 pooling, so 266 of 280 soft tokens, and the tower and the caller agree there. Only the stand-in device disagrees, because nothing computes the mask on it. Asking a shipped model to carry a host-side guard for that would be paying for a path nobody runs, and it would leave the real question — whether the tower and the caller agree on a device — permanently answered by construction rather than by observation.

So the harness that chose the stand-in device now corrects for it, scoped to dummy-device runs and a no-op on anything already stripped. The arithmetic and its tests moved there intact.

What stays worth saying: the general shape is that host code whose control flow reads a value out of a device output cannot be decided on a device that does not compute, and it is usually a smell because the host already knew the answer. If we ever want the mask derived from the position ids in optimum for its own sake — one source instead of two — that is a separate argument to make on its merits, not as a CI fix.

PR #679, opened 2026-08-20
other(auto): recover concrete config class for bare PretrainedConfig inputs

rebel-thkim:
## Type of Change

## Changes Overview
- `auto_factory.py`: add `_upgrade_bare_config`, which rebuilds a bare base-class `PretrainedConfig` as its concrete config class (e.g. `MistralConfig`) by resolving `config.architectures[0]` against the `transformers` namespace and re-instantiating via that class's `from_dict`. The parser-supplied `model_type` is dropped so the concrete class's canonical value wins.
- Hook it in two places: `infer_hf_model_class` (so the exact-config-class model lookup hits) and `_BaseAutoModelClass.from_pretrained` (so the upgraded config — not the bare one — propagates to the resolved class's `from_pretrained` and eager model loading).
- Unit tests covering RBLN class resolution from a bare config, config-class recovery/equivalence, and the leave-unchanged fallback for unresolvable inputs.

## Motivation and Context
`mistralai/Mistral-7B-Instruct-v0.3` fails at vLLM engine startup on the optimum-rbln path. vLLM auto-selects the Mistral config format (`params.json`) for repos shipping `consolidated*.safetensors`, and its parser returns a **base `PretrainedConfig`** instance. The auto factory resolves model classes by exact config class, so `infer_hf_model_class` raises `Unrecognized configuration class ... PreTrainedConfig`.

Two subtleties found while verifying against the real v0.3 repo:
- The parsed config's `model_type` is `"transformer"`, not `"mistral"`, so a `CONFIG_MAPPING[model_type]`-based recovery would also fail. `architectures` (`["MistralForCausalLM"]`) is the reliable identity and is always set by vLLM's adapter.
- Fixing class resolution alone is not enough: the bare config flows through `_export → get_pytorch_model → MistralForCausalLM.from_pretrained(..., config=...)` and crashes there (`AttributeError: 'PreTrainedConfig' object has no attribute 'pad_token_id'`). Hence the propagation hook in `from_pretrained`.

Verified with the real v0.3 config: the upgraded config is field-for-field identical to the `config.json`-loaded `MistralConfig`, resolves to `RBLNMistralForCausalLM`, and constructs the eager model (meta device).

## Related Issues
rebellions-sw/vllm-optimum-executor#76

## Test
- `pytest tests/test_base.py -k "bare or upgrade"` → 2 passed
- `ruff check` / `ruff format --check` clean

### DIFF
--- a/src/optimum/rbln/transformers/models/auto/auto_factory.py
+++ b/src/optimum/rbln/transformers/models/auto/auto_factory.py
@@ -17,6 +17,7 @@ import warnings
 from pathlib import Path
 from typing import Any
 
+import transformers
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel
 from transformers.dynamic_module_utils import get_class_from_dynamic_module
 from transformers.models.auto.auto_factory import _get_model_class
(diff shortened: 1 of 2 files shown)


rebel-eunji (commented):
lgtm 👍

rebel-eunji (approved):

src/optimum/rbln/transformers/models/auto/auto_factory.py
```diff
@@ -31,6 +32,28 @@
...
+    Non-HF config parsers (e.g. vLLM's mistral-format loader) hand over an
+    instance of the base ``PretrainedConfig``, whose class carries no model
+    identity and whose ``model_type`` can be a parser artifact ("transformer").
+    Recover the concrete class from ``architectures`` so class-based model
+    resolution and eager model loading work downstream. Returns the config
+    unchanged if the architecture cannot be resolved.
+    """
+    architectures = getattr(config, "architectures", None) or []
+    hf_model_cls = getattr(transformers, architectures[0], None) if architectures else None
+    config_cls = getattr(hf_model_cls, "config_class", None)
```
rebel-eunji, line 47: nit: Too many guards.

PR #267, opened 2025-08-29
model: Refactor multimodal

rebel-kblee:
# Pull Request Description
- mem optimization
- architecting -> lazy import for submodule config
- control rbln_config
- padding logic if needs

## Type of Change

- [ ] New Model Support
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please describe):

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works (If needed)

## Related Issues

----
# Conventional commit
```
type(optional scope): description
```
----
# Type candidate
  - Model Updates
    - `model`: Adding New models or Bugfix for existing models
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
    - `performance`: Optimizing some models or this library itself
      - ex) Loading RBLNModel faster
      - ex) Optimizing Memory Usage of DecoderOnlyModel
  - Code Refactor
    - `refactor`: Re-arrange class architecture, or more.
      - ex) Refactor Seq2Seq
  - Documentation
    - `doc`: Update docstring only     
  - Library Dependencies
    - `dependency`: Update requirements, something like that.
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

PR #681, opened 2026-08-24
model: propagate top-level batch_size to VLM language-model submodules

rebel-thkim:
Follow-up to #650. That PR fixed the vllm-rbln case (`{"language_model": {"batch_size": N}}`, no top-level key) by dropping the `batch_size` kwarg from gemma4's `initialize_submodule_config` call — but that also removed the only path by which a **top-level** `batch_size` reached the language model, so `rbln_config={"batch_size": 2}` silently compiled the LM at batch 1. gemma3 / llava / llava_next / blip_2 / idefics3 had the same gap all along (their parent docstrings say "The batch size for inference", but the value never left the parent config).

## Fix

Propagate the top-level `batch_size` to the language-model submodule as a **soft default**, the way paligemma already does: pass `batch_size=batch_size` **without** `force_kwargs`. In `initialize_submodule_config` the submodule dict is applied on top of the parent's kwargs, so:

- top-level only → propagates to the language model
- submodule only → submodule value used (vllm-rbln style, the #650 case — unchanged)
- both set → submodule wins (dev behavior, unchanged)

The parent→submodule kwargs machinery in `initialize_submodule_config` exists to propagate infra settings — `num_devices` (TP), `npu`, and the runtime options — down to submodules; `force_kwargs` came in with it (#267) for literal vision pins (`batch_size=1`). Reusing it to force the parent's *resolved* `batch_size` onto the language model overloaded that mechanism and turned a default into a hard requirement — which is why #650 was needed in the first place. `force_kwargs` remains only where a literal pin is intended (vision towers at `batch_size=1`).

Applied to gemma4 (restores the propagation removed in #650, without the conflict), gemma3, llava, llava_next, blip_2, and idefics3 (`text_model`). The qwen-family VLMs need nothing — their parent config *is* the CausalLM config.

**colqwen2** had a live instance of the same wrong `force_kwargs` usage: it forces `batch_size` / `output_hidden_states` that may be `None`, so `vlm={"batch_size": 4}` alone raised `Parameter conflict: 4 != None`. Dropping `force_kwargs` from its `vlm` call fixes it the same way.

## Before / after

Idefics3 (`trl-internal-testing/tiny-Idefics3ForConditionalGeneration`), verified by compiling on this branch vs `dev` (2a13f24f):

**model_0** — `rbln_config={"batch_size": 2, "text_model": {"max_seq_len": 2048}}`

| | parent | text_model |
|---|---|---|
| before | 2 | **1** (top-level silently ignored) |
| after | 2 | **2** |

**model_1** — `rbln_config={"text_model": {"batch_size": 2, "max_seq_len": 2048}}` (vllm-rbln style)

| | parent | text_model |
|---|---|---|
| before | 1 | 2 |
| after | 1 | 2 (unchanged) |

**model_2** — `rbln_config={"batch_size": 1, "text_model": {"batch_size": 2, "max_seq_len": 2048}}`

| | parent | text_model |
|---|---|---|
| before | 1 | 2 (submodule wins) |
| after | 1 | 2 (unchanged) |

Only model_0 changes; the other two keep their dev behavior.

## Side effects checked

- Reload of saved artifacts is unaffected: `RBLNModelConfig.from_pretrained` pre-instantiates submodule dicts before calling the parent ctor, and instances bypass `initialize_submodule_config` merging entirely. Verified end-to-end (compile → `save_pretrained` → `from_pretrained`) with the model_1-style artifact (saved parent `batch_size=1`, `text_model.batch_size=2`).
- vllm-rbln builds submodule-only dicts (`{"language_model"/"text_model": {...}}`, no top-level `batch_size`) — unaffected.
- Vision towers stay pinned: an explicit `vision_tower={"batch_size": 2}` still raises as before.

## Tests

`tests/test_config.py`:
- `test_composite_vlm_batch_size_propagation` — parametrized over the six composite configs: top-level-only propagates, submodule-only works, submodule wins on both-set, equal both-set works (config level, no compile)
- `test_colqwen2_submodule_only_kwargs_no_conflict`

### DIFF
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -543,6 +543,10 @@ class RBLNModelConfig(RBLNSerializableConfigProtocol):
             return submodule_config
 
         if isinstance(submodule_config, dict):
+            # A None-valued kwarg means the parent left it unset: it neither propagates to the
+            # submodule nor participates in the force_kwargs conflict check.
+            kwargs = {key: value for key, value in kwargs.items() if value is not None}
+
             from_predecessor = self._runtime_options.copy()
             from_predecessor.update(
                 {
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -543,10 +543,6 @@ class RBLNModelConfig(RBLNSerializableConfigProtocol):
             return submodule_config
 
         if isinstance(submodule_config, dict):
-            # A None-valued kwarg means the parent left it unset: it neither propagates to the
-            # submodule nor participates in the force_kwargs conflict check.
-            kwargs = {key: value for key, value in kwargs.items() if value is not None}
-
             from_predecessor = self._runtime_options.copy()
             from_predecessor.update(
                 {
--- a/src/optimum/rbln/transformers/models/blip_2/configuration_blip_2.py
+++ b/src/optimum/rbln/transformers/models/blip_2/configuration_blip_2.py
@@ -117,4 +117,6 @@ class RBLNBlip2ForConditionalGenerationConfig(RBLNModelConfig):
             submodule_config=vision_model, batch_size=1, force_kwargs=True
         )
         self.qformer = self.initialize_submodule_config(submodule_config=qformer, batch_size=1, force_kwargs=True)
-        self.language_model = self.initialize_submodule_config(submodule_config=language_model)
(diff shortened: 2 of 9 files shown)


rebel-dkhong (approved):

rebel-seinpark:
@rebel-thkim How about add `batch_size=1, force_kwargs=True` explicitly in `configuration_qwen*vl` series?
If someone sets `rbln_config = {visual: {batch_size:1}}`, it is ignored silently.

PR #682, opened 2026-08-24
model(qwen-vl): align vision encoder batch_size handling with qwen3.5

rebel-seinpark:
## Summary

The Qwen2-VL, Qwen2.5-VL and Qwen3-VL vision encoders all run one image at a
time (per-image loop), so their batch dimension is always 1. Previously this
invariant was left implicit: the vision configs did not expose `batch_size` and
the vision `_update_rbln_config` hardcoded a leading `1` in the compile input
shapes.

This PR generalizes the vision batch handling to the **same form as qwen3.5**:

- Expose an explicit `batch_size: int = 1` on the vision configs
  (`RBLNQwen2VisionTransformerPretrainedModelConfig`,
  `RBLNQwen2_5_VisionTransformerPretrainedModelConfig`,
  `RBLNQwen3VLVisionModelConfig`) with a guard that raises if it is not 1.
- Force it from the parent configs (both `ForConditionalGeneration` and `Model`
  variants) via
  `initialize_submodule_config(submodule_config=visual, batch_size=1, force_kwargs=True)`,
  so a conflicting submodule value fails fast with a clear error instead of the
  previous confusing "Unexpected arguments".
- Read `rbln_config.batch_size` in the vision `_update_rbln_config` instead of
  hardcoding `1` for the `attn_mask` / `full_attn_masks` / `cos` / `sin` leading
  dimension (the `window_attn_masks` leading dim is the window count, so it is
  left unchanged).

Behavior is unchanged (vision batch is still 1); this only makes the invariant
explicit, guarded, and consistent with qwen3.5.

## Backward compatibility

Loading models compiled before this change is unaffected: the saved vision
config simply lacks `batch_size`, so it falls back to the default `1`; on load
`initialize_submodule_config` short-circuits for already-instantiated configs
(no `force_kwargs` check), and `_update_rbln_config` is not re-run, so the
compiled graph shapes are read as-is from the saved compile configs.

## Test plan

- [ ] Compile Qwen2-VL / Qwen2.5-VL / Qwen3-VL and confirm graphs and runtime are unchanged
- [ ] Load a model compiled on `dev` (pre-change) and confirm it still loads and runs
- [ ] Passing `visual={"batch_size": >1}` raises a clear `batch_size=1` error

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/optimum/rbln/transformers/models/qwen2_vl/configuration_qwen2_vl.py
+++ b/src/optimum/rbln/transformers/models/qwen2_vl/configuration_qwen2_vl.py
@@ -45,7 +45,7 @@ class RBLNQwen2VLForConditionalGenerationConfig(RBLNDecoderOnlyModelForCausalLMC
                 "RBLNQwen2VLForConditionalGenerationConfig does not allow `use_inputs_embeds` to be set to False, "
                 "as RBLNQwen2VLForConditionalGeneration accepts only `inputs_embeds` as input."
             )
-        self.visual = self.initialize_submodule_config(submodule_config=visual)
+        self.visual = self.initialize_submodule_config(submodule_config=visual, batch_size=1, force_kwargs=True)
 
 
 class RBLNQwen2VLModelConfig(RBLNDecoderOnlyModelConfig):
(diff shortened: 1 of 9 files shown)


src/optimum/rbln/transformers/models/qwen2_vl/configuration_qwen2_vl.py
```diff
@@ -57,11 +57,11 @@ class RBLNQwen2VLModelConfig(RBLNDecoderOnlyModelConfig):
 
     def __init__(self, visual: RBLNModelConfig | None = None, **kwargs: dict[str, Any]):
         super().__init__(**kwargs)
-        self.visual = self.initialize_submodule_config(submodule_config=visual)
+        self.visual = self.initialize_submodule_config(submodule_config=visual, batch_size=1, force_kwargs=True)
 
 
 class RBLNQwen2VisionTransformerPretrainedModelConfig(RBLNModelConfig):
-    def __init__(self, max_seq_len: int | list[int] = None, **kwargs: dict[str, Any]):
+    def __init__(self, max_seq_len: int | list[int] = None, batch_size: int = 1, **kwargs: dict[str, Any]):
```
rebel-thkim: I think we'll need to add the `deprecate_kwargs` decorator.

rebel-thkim (approved):
👍

PR #685, opened 2026-08-24
other(tests): serve HF loads from the local cache before calling the Hub

rebel-thkim:
## What
Mitigation for the HF rate-limit incident (option 1: code-level cache-first fallback).

Wraps every test-side `from_pretrained` call that takes a hub model id in a `from_pretrained_cache_first()` helper:

```python
try:
    return loader.from_pretrained(model_id, local_files_only=True, **kwargs)
except OSError:
    return loader.from_pretrained(model_id, **kwargs)
```

- With a warm cache, tests make **zero** Hub API calls (no etag HEAD requests at all).
- On a cache miss (e.g. a PR adding a new model), it falls back to the normal online download.
- Also moves the `test_llm.py` tokenizer cache from the instance to the concrete class: unittest builds a new instance per test method, so the tokenizer was re-resolved against the Hub on every single test.

## Why
Even with a warm HF_HOME cache, `from_pretrained` fires an etag HEAD request per resolved file, and those count against the account-wide HF rate limit (~94 test classes x several requests each, plus the per-method tokenizer reloads = hundreds to ~1k requests per PR CI run). Since the account is shared with the vllm-rbln and compiler CIs, this was a major contributor to the recent rate-limit incident.

## Verification
- `pytest --collect-only`: all 5 test modules import, 778 tests collected, no errors
- Helper smoke test: cache-hit path (offline reload succeeds) and miss path (OSError, then online fallback) both verified
- `ruff check` / `ruff format` pass

Related: the CI-level two-phase approach (cache warming + `HF_HUB_OFFLINE=1`) is proposed separately in #686. Only one of the two should be adopted.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -102,6 +102,15 @@ def test_deprecate_method_raises_at_or_past_cutoff(current_version, expect_raise
         stub()
 
 
+def from_pretrained_cache_first(loader, model_id, **kwargs):
+    # Serve from the local HF cache without any Hub API call (CI shares one
+    # account-level rate limit); go online only on a cache miss.
+    try:
+        return loader.from_pretrained(model_id, local_files_only=True, **kwargs)
+    except OSError:
+        return loader.from_pretrained(model_id, **kwargs)
+
+
 DUMMY_DEVICE_CODE = -1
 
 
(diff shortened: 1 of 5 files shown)


rebel-junamsong (approved):
LGTM

PR #698, opened 2026-08-27
model(decoderonly): stop dropping logits_to_keep=0

rebel-eunji:
## Type of Change

- [x] Bug fix

## Overview

`RBLNDecoderOnlyModelConfig` used `or` to resolve `logits_to_keep`:

```python
self.logits_to_keep = logits_to_keep or self._default_logits_to_keep
```

Because `0` is falsy, it was incorrectly replaced by the class default. This prevented requesting logits for all positions and caused saved configurations with `0` to reload as `1`.

The fallback now applies only when the value is `None`:

```python
self.logits_to_keep = logits_to_keep if logits_to_keep is not None else self._default_logits_to_keep
```

A regression test verifies that:

- `logits_to_keep=0` is preserved during construction.
- `0` survives a save/reload round trip.
- An unset value still uses the class default.

## Verification

- `tests/test_config.py`: 39 passed, 2 skipped
- `ruff format --check`: passed
- `ruff check`: passed

### DIFF
--- a/src/optimum/rbln/transformers/models/decoderonly/configuration_decoderonly.py
+++ b/src/optimum/rbln/transformers/models/decoderonly/configuration_decoderonly.py
@@ -247,7 +247,9 @@ class RBLNDecoderOnlyModelConfig(RBLNModelConfig):
         if phases is not None:
             self.validate_phases_type(phases)
         self.phases = phases or self._default_phases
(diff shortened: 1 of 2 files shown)


rebel-jongho (approved):
This is a great catch. 

nit: Could you remove comments & tests in the code?

PR #699, opened 2026-08-27
model(vlm): handle attention_mask=None in raw forward() calls

rebel-seinpark:
## Type of Change

## Changes Overview

Added raw-call defaults at the `forward()` entry points of four Qwen VLMs (qwen2_vl, qwen2_5_vl, qwen3_vl, qwen3_5):

- `attention_mask=None` → `torch.ones_like(input_ids)` (HF semantics: no mask means every token is attended)
- `generate_idx=None` → `attention_mask.sum(dim=-1, keepdim=True).int()` (same default as `RBLNDecoderOnlyModelForCausalLM.forward`)

blip_2 / gemma4 / exaone4_5 already handle None and are unchanged.

## Motivation and Context

The Qwen VLM implementations index `attention_mask` (and `generate_idx`) without a None guard (`attention_mask[b_idx]`, `generate_idx[b_idx]`), assuming they are always provided. That assumption only holds on the `generate()` path, where `GenerationMixin._prepare_attention_mask_for_generation` synthesizes a mask.

Calling `forward()` directly bypasses that safety net and crashes with `TypeError: 'NoneType' object is not subscriptable`. Real occurrences:

- **Any pipeline that uses a Qwen model as its text encoder** (e.g. Cosmos-Predict2.5): calls raw forward without a mask, like `text_encoder(input_ids, output_hidden_states=True)` → crash. The original HF models accept this call (None means every token is attended; the causal mask is always applied inside the model), so this is a behavioral mismatch with upstream.
- **RBLNColQwen2**: explicitly allows `attention_mask=None` and passes it straight into the qwen2_vl backbone forward → a live, reproducible bug today.

The fix inserts the same defaults at the same points where HF handles None, and is a no-op for the existing `generate()` path where a mask is always supplied. Note that HF does the same: `get_rope_index` skips padding-stripping via an `if attention_mask is not None:` guard, and forward runs causal-only when no mask is given.

### DIFF
--- a/src/optimum/rbln/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
+++ b/src/optimum/rbln/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
@@ -547,6 +547,8 @@ class RBLNQwen2_5_VLModel(RBLNDecoderOnlyModel):
         mm_token_type_ids: torch.IntTensor | None = None,
         **kwargs,
     ) -> RBLNDecoderOnlyOutput:
+        if attention_mask is None:
+            attention_mask = torch.ones_like(input_ids)
         inputs_embeds, position_embed, rope_deltas = self._preprocess_prefill(
             input_ids,
             attention_mask,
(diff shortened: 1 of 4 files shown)


rebel-thkim (approved):
:+1

PR #701, opened 2026-08-27
model(vlm): resolve text config for output_hidden_states buffers

rebel-seinpark:
> **⚠️ Branch Target**: `dev`

## Type of Change

## Changes Overview

- `RBLNRuntimeModel._prepare_prefill_outputs`: size the hidden-state output buffers via `self.config.get_text_config().hidden_size / num_hidden_layers` instead of flat `self.config.*` access. `get_text_config()` returns the config itself for plain LLMs, so this is a no-op for text-only models — and the same function was already used one branch above to resolve `logits_last_dim`.
- `RBLNQwen2_5_VLModel`: the hidden-state copy loops iterated `self.config.num_hidden_layers`; use `self.config.text_config.num_hidden_layers`, matching what the sibling implementations (qwen2_vl, qwen3_vl, qwen3_5) already do.
- Add `TestQwen2_5_VLForConditionalGeneration_OutputHiddenStates`, mirroring the existing qwen3_vl output_hidden_states test (batch of two left-padded prompts, `output_hidden_states=True`).

## Motivation and Context

transformers 5.x composite VLM configs (e.g. `Qwen2_5_VLConfig`) no longer expose text attributes like `hidden_size` / `num_hidden_layers` at the top level — they live in `config.text_config`. Two spots still assumed a flat config, so running Qwen2.5-VL with `output_hidden_states=True` crashed:

```
AttributeError: 'Qwen2_5_VLConfig' object has no attribute 'hidden_size'
```

The path was never exercised before: `generate()` does not request hidden states per forward step, and qwen3_vl sidesteps the runtime half by passing `config.text_config` into its runtime. Using a VLM as a text encoder — e.g. Cosmos-Predict2.5 calling `text_encoder(input_ids, output_hidden_states=True)` — hits both spots.

qwen2_5_vl had no output_hidden_states test (qwen3_vl does), so this PR adds one to cover the regression.

### DIFF
--- a/src/optimum/rbln/transformers/models/decoderonly/decoderonly_runtime_utils.py
+++ b/src/optimum/rbln/transformers/models/decoderonly/decoderonly_runtime_utils.py
@@ -452,8 +452,8 @@ class RBLNRuntimeModel(RBLNPytorchRuntime):
             int(torch.nonzero(attention_mask, as_tuple=False)[0][0].item()) if attention_mask is not None else 0
         )
 
+        text_config = self.config.get_text_config()
         if self.logits_last_dim is None:
-            text_config = self.config.get_text_config()
             logits_last_dim = text_config.vocab_size if self.rbln_config.can_generate else text_config.hidden_size
         else:
             logits_last_dim = self.logits_last_dim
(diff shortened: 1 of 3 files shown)


rebel-thkim (approved):
👍

rebel-kblee (approved):

PR #702, opened 2026-08-27
other(config): keep the load-time object gate, reduce partial configs at the call site

rebel-seinpark:

## Type of Change
- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Overview

**The load-time contract for `rbln_config` objects is kept strict; the one caller that legitimately holds a partial object converts it to a dict at the call site.**

- `RBLNModelConfig.get_load_overrides()` (new): extracts everything that must cross the load boundary — the explicitly-set **runtime options** (`device`, `device_map`, `create_runtimes`, `activate_profiler`, `timeout`) and the **`subclass_non_save_attributes`** (load-behavior flags such as `_load_visual_runtime`, which are never serialized, so they can only travel with the caller) — recursively over submodules. Compile-time attributes are excluded: the artifact's `rbln_config.json` is the source of truth for them. A submodule left as a plain dict contributes runtime options only, since its config class (and thus its non-save attribute list) is unknown at that point.
- `RBLNDiffusionMixin.from_pretrained` (load path): the pipeline config is objectified from user input, so the submodule config it passes down is a **default-filled partial**, not the complete config recorded in the artifact. It is now reduced with `get_load_overrides()` and goes through the existing, validated dict-merge path:

```
load rbln_config.json from disk        <- base (the complete config, with all compile-time info)
        +
get_load_overrides()                   <- {"device": [...], "_load_visual_runtime": false, "visual": {"device": ...}}
        v  merge (dict path)
final config
```

- `RBLNModelConfig.from_pretrained`: the object branch (the "gate") is **kept**. Only a config loaded from the same artifact can match it — `_compile_cfgs` exists nowhere else — so any hand-constructed object is rejected instead of being silently accepted or silently ignored. The error message now explains this and points at the dict path / `get_load_overrides()`. (An earlier iteration of this PR reduced every object to overrides at this entry point; that turned misuse into silence for *all* load paths and was reverted in favor of the call-site conversion.)
- Fix the kwarg-collision guard to check the `rbln_`-prefixed key, preserving the existing precedence where explicit `rbln_*` kwargs win over the config bundle.

## Motivation and Context

When a diffusers pipeline loads compiled models (`export=False`), `RBLNDiffusionMixin.from_pretrained` materializes the user's `rbln_config` dict into a pipeline config **object**, and while loading each submodule it passes `getattr(rbln_config, submodule_name)` — the objectified submodule config — straight into the submodule's `from_pretrained(rbln_config=...)`.

This objectification cannot produce a complete config:

- Every field the user did not provide is **filled with defaults**. Information fixed at compile time (`max_seq_len`, `_compile_cfgs`, ...) exists only in the artifact's `rbln_config.json`.
- A nested submodule (e.g. `visual` of Qwen2.5-VL) cannot even determine its config class at this point, so it **remains a plain dict**.

Meanwhile, the object branch of `RBLNModelConfig.from_pretrained` compares each submodule against the config loaded from disk. That comparison can only ever pass for a round-trip (a config loaded from the same artifact, with runtime options freely mutated — runtime options are excluded from the comparison). A default-filled partial can never pass it, so loading any model with nested submodules through a pipeline always failed:

```
ValueError: Passed rbln_config has different attributes for submodule visual than the config_file
```

Found with Cosmos-Predict2.5 (Qwen2.5-VL text encoder). The fix keeps the gate — it is the only thing that rejects a hand-constructed object carrying wrong compile-time values, and everything it blocks *should* be blocked — and converts the partial to a dict exactly where the partial is created.

The extraction covering `subclass_non_save_attributes` is load-bearing, not cosmetic: a runtime-options-only reduction silently drops flags like `_load_visual_runtime=False` at the pipeline→submodule boundary (they are non-save, so the disk config cannot restore them either), which re-loads a module the pipeline intended to skip.

## Related Issues

### DIFF
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -817,6 +817,47 @@ class RBLNModelConfig(RBLNSerializableConfigProtocol):
 
         return serializable_map
 
+    def get_runtime_overrides(self) -> dict[str, Any]:
+        """
+        Extract the runtime options set on this config, for passing to from_pretrained.
+
+        Returns only the runtime options (device, device_map, create_runtimes,
+        activate_profiler, timeout) that are explicitly set, recursively including
+        submodules. Compile-time attributes are not included: when loading a compiled
+        model they come from the artifact's rbln_config.json, and this dict is merged
+        on top of it.
+
+        Returns:
+            Dictionary of explicitly-set runtime options, keyed like the rbln_config
+            dict accepted by from_pretrained.
+        """
+
+        def filter_runtime(cfg):
+            # Recursively extract runtime options from config (RBLNModelConfig or dict)
+            if isinstance(cfg, RBLNModelConfig):
+                return cfg.get_runtime_overrides()
+            if not isinstance(cfg, dict):
+                return None
+            result = {}
+            for k, v in cfg.items():
+                if k in RUNTIME_KEYWORDS:
+                    if v is not None:
+                        result[k] = v
+                elif isinstance(v, dict):
+                    nested = filter_runtime(v)
+                    if nested:
+                        result[k] = nested
+            return result or None
+
+        result = {k: v for k, v in self._runtime_options.items() if v is not None}
+
+        for name in self.submodules:
+            filtered = filter_runtime(getattr(self, name, None))
+            if filtered:
+                result[name] = filtered
+
+        return result
+
     def __repr__(self):
         repr_dict = self._prepare_for_serialization()
         return json.dumps(repr_dict, indent=2)
@@ -917,7 +958,7 @@ class RBLNModelConfig(RBLNSerializableConfigProtocol):
             # At load time, a passed config object is a carrier of runtime overrides,
             # not a complete config. Extract its runtime options (top-level and per-submodule)
             # and merge them through the dict path below.
-            rbln_config = rbln_config.to_dict()
+            rbln_config = rbln_config.get_runtime_overrides()
 
         if isinstance(rbln_config, dict):
             for key, value in rbln_config.items():
--- a/src/optimum/rbln/configuration_utils.py
+++ b/src/optimum/rbln/configuration_utils.py
@@ -955,7 +955,7 @@ class RBLNModelConfig(RBLNSerializableConfigProtocol):
             logger.warning(f"Expected {cls.__name__}, but got {cls_reserved.__name__}.")
 
         if isinstance(rbln_config, RBLNModelConfig):
-            # At load time, a passed config object is a carrier of runtime overrides,
+            # At `export=False` time, a passed config object is a carrier of runtime overrides,
             # not a complete config. Extract its runtime options (top-level and per-submodule)
             # and merge them through the dict path below.
             rbln_config = rbln_config.get_runtime_overrides()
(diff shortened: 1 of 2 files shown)


src/optimum/rbln/configuration_utils.py
```diff
@@ -913,9 +954,15 @@ def from_pretrained(
         if cls_reserved != cls:
             logger.warning(f"Expected {cls.__name__}, but got {cls_reserved.__name__}.")
 
+        if isinstance(rbln_config, RBLNModelConfig):
+            # At load time, a passed config object is a carrier of runtime overrides,
+            # not a complete config. Extract its runtime options (top-level and per-submodule)
+            # and merge them through the dict path below.
```
rebel-seinpark:
```suggestion
            # At `export=False` time, a passed config object is a carrier of runtime overrides,
            # not a complete config. Extract its runtime options (top-level and per-submodule)
            # and merge them through the dict path below.
```

src/optimum/rbln/configuration_utils.py
```diff
@@ -943,11 +993,20 @@ def from_pretrained(
         if isinstance(rbln_config, RBLNModelConfig):
             config_file.update(rbln_config._runtime_options)
 
-            # update submodule runtime
+            # A config object passed at load time must carry the same compile-time attributes as
+            # the artifact, which only a config loaded from that artifact can (`_compile_cfgs`
+            # exists nowhere else). The comparison excludes runtime options and other non-save
+            # attributes, since `__repr__` serializes neither.
```
rebel-seinpark:
```suggestion
            # update submodule runtime
```

src/optimum/rbln/configuration_utils.py
```diff
@@ -943,11 +993,20 @@ def from_pretrained(
...
                 if str(config_file[submodule]) != str(getattr(rbln_config, submodule)):
                     raise ValueError(
-                        f"Passed rbln_config has different attributes for submodule {submodule} than the config_file"
+                        f"The `rbln_config` object passed at load time does not match the artifact's "
+                        f"saved config for submodule `{submodule}`. Only a config loaded from the same "
+                        f"artifact (e.g. via `{cls.__name__}.from_pretrained`) can be passed as an "
+                        f"object, since compile-time attributes exist only in the artifact's "
+                        f"rbln_config.json. To override runtime options, pass a dict instead, e.g. "
+                        f"`rbln_config={{'{submodule}': {{'device': 0}}}}`, or reduce the object with "
+                        f"`get_load_overrides()`."
```
rebel-seinpark:
```suggestion
                        f"Passed rbln_config has different attributes for submodule {submodule} than the config_file"
```

src/optimum/rbln/diffusers/modeling_diffusers.py
```diff
@@ -233,6 +233,13 @@ def from_pretrained(
 
                     submodule_cls = get_rbln_model_cls(class_name)
                     submodule_config = getattr(rbln_config, submodule_name)
+                    if isinstance(submodule_config, RBLNModelConfig):
+                        # The pipeline config above was objectified from user input, so this
+                        # submodule config is a default-filled partial, not the complete config
+                        # recorded in the artifact. Reduce it to its load overrides (runtime
+                        # options and load-behavior flags) and let the artifact's
+                        # rbln_config.json fill in the compile-time attributes.
```
rebel-seinpark:
```suggestion
                        # This submodule config is a default-filled partial, not the complete
                        # config recorded in the artifact.
```

rebel-thkim (approved):
👍

PR #710, opened 2026-09-03
model(qwen3.5): restore HF fp32 casts in the GatedDeltaNet kernel

rebel-seinpark:
## Summary

Restores the fp32 compute that HF's `torch_chunk_gated_delta_rule` / `torch_recurrent_gated_delta_rule`
use inside the GatedDeltaNet kernel, which the RBLN rewrite had intentionally dropped. The kernel now
matches HF semantics:

- **Chunk (prefill) rule**: upcast the inputs (`q/k/v/beta/g`) and the carried `initial_state` to fp32;
  cast the outputs (`core`, `last_recurrent_state`) back to the graph dtype at the kernel boundary.
- **Single-step (decode) rule**: same input/state upcast (outputs were already cast back).
- **Gate**: `g = -A_log.float().exp() * softplus(a.float() + dt_bias)` — per HF's comment, this guards
  `A_log.exp()` from overflowing to `-inf` when the model runs in fp16 (bf16 shares fp32's exponent
  range, so the default bf16 path was never at risk).

## Why now

The casts were originally removed to keep the kernel in a single device dtype. A device experiment shows
the fp32 island is safe to restore, so we take the HF-faithful form:

- **Compiles**: the fp32 island — including fp32 `rbln::tri_recur_update` — lowers without issues.
- **Parity on par** (4-layer real Qwen3.5-0.8B, teacher-forced 30 steps, flash attn):

| | single-dtype (before) | fp32 kernel (this PR) |
| --- | --- | --- |
| logit pearson mean | 0.99985 | 0.99985 |
| logit pearson min | 0.99957 | 0.99963 |
| argmax match | 29/30 | 28/30 (extra flip is a near-tie, HF top1-top2 gap 0.125) |

The default path loads the checkpoint in bf16 (transformers v5 `dtype="auto"`), where the upcast adds no
information — parity is unchanged, as expected. The gain is HF fidelity and the fp16-load guard.

## Testing

- Device parity: 4-layer real-weight teacher-forced run (above).
- `ruff check` / `ruff format` clean.

### DIFF
--- a/src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
+++ b/src/optimum/rbln/transformers/models/qwen3_5/qwen3_5_architecture.py
@@ -142,7 +142,7 @@ def rbln_chunk_gated_delta_rule(
         query = l2norm(query, dim=-1, eps=1e-6)
         key = l2norm(key, dim=-1, eps=1e-6)
     query, key, value, beta, g = [
-        x.transpose(1, 2).contiguous().to(initial_dtype) for x in (query, key, value, beta, g)
+        x.transpose(1, 2).contiguous().to(torch.float32) for x in (query, key, value, beta, g)
     ]
     scale = 1 / (query.shape[-1] ** 0.5)
     query = query * scale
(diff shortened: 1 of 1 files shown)


rebel-junamsong (approved):
LGTM

PR #714, opened 2026-09-07
model: moe drop redundant expert weight copies and release HF fused gate_up_proj

rebel-thkim:
## Type of Change
- [x] Performance improvement
- [x] Code refactoring

## Changes Overview
HF 5.x stores MoE experts as fused `gate_up_proj [E, 2I, H]` and `down_proj [E, H, I]`, while
`custom_moe_glu` takes gate / up / down separately. The wrappers (qwen2_moe, qwen3_moe, qwen3_vl_moe,
mixtral, gemma4) split the fused tensor at construction, and did so with several redundant copies per
layer that stayed alive next to the HF model during compile. One commit per step:

1. **Drop redundant copies** (07548e06). `gate_up_proj.detach().clone()` before slicing (the slice's
   `.contiguous()` already copies), `down_proj.detach().clone().contiguous()` (already contiguous in the
   op's layout), the cloned router `gate.weight`, and full-size `nn.Linear(H, E*I)` weights that were
   randomly initialized only to be overwritten. Only the two gate/up halves are copied now; `down_proj`
   and the router weight share storage with the HF model.
2. **Release HF `gate_up_proj` after the split** (004a1327). The HF experts module is not executed after
   wrapping, so the split halves become the only copy and the per-model footprint stays flat.
3. **GPT-OSS** (68c63d23). The even/odd mxfp4 slices for gate/up blocks, scales and bias were
   non-contiguous views registered as buffers. The compiler copies them at ingest while the HF originals
   are alive, and `normalize_contiguous_` (#600) only walks parameters, so the same doubling happened at
   compile time. Copy once in the wrapper and release the HF `gate_up_proj` (`gate_up_proj_blocks` on
   older layouts). `down_proj` was already a contiguous view and stays shared.
4. **Register the split tensors as buffers** (0b332b10). The `nn.Linear(1, 1)` shells were never called;
   forward only read `.weight`. Register `gate_proj_weight` / `up_proj_weight` / `down_proj_weight`
   directly, as GPT-OSS does. rebel treats parameters and buffers alike
   (`named_parameters() + named_buffers()`). mixtral already had bare parameters and is unchanged here.

## Motivation and Context
Compiling Qwen3-30B-A3B in fp32 (~120 GB of weights) exceeds the 175 GB host budget because the
wrapper held a second copy of every expert weight while the HF model was still alive.

Measured on one real `Qwen3MoeSparseMoeBlock` (E=64, I=768, H=2048, fp32; HF expert weights = 1150 MB),
wrapping it with `Qwen3MoeSparseMoeBlock` from this repo and compiling that single block with
`rebel.compile_from_torch` (input `[1, 16, H]`):

| commit | wrap peak | steady after wrap | compile peak | HF `gate_up_proj` alive |
|---|---|---|---|---|
| dev (9eb38f94) | +267% | +100% | 4217 MB (3.67x) | yes |
| 1. drop copies | +66% | +66% | 3564 MB (3.09x) | yes |
| 2. release HF fused | +66% | ±0% | 2794 MB (2.42x) | no |
| 4. buffers (final) | +67% | ±0% | 2792 MB (2.44x) | no |

- *wrap peak*: maximum extra RSS at any moment during wrapper construction, relative to the HF expert weights.
- *steady*: extra RSS that remains after construction (HF model still alive). This is what accumulates
  across all layers of a model.
- *compile peak*: process peak RSS through `compile_from_torch` of the wrapped block, absolute and relative
  to the HF expert weights.

The remaining +66% transient is per layer (wrappers are built layer by layer), so it is negligible for a
full model. The remaining compile-time overhead (rebel-compiler's nn.Module clone / constant export) is
unchanged by this PR and is being looked at separately.

Passing strided views of `gate_up_proj` straight to the op was also tried: it compiles and gives
bit-identical output, but the compiler materializes contiguous copies anyway, so it saves nothing over
commit 1. A fused-weight variant of `custom_moe_glu` would only remove the per-layer transient and was not
pursued.

### Behavior change
After wrapping, the *source* HF model's `experts.gate_up_proj` (GPT-OSS: `gate_up_proj` /
`gate_up_proj_blocks`) is `None`, so its `forward` no longer works. `from_pretrained` builds the HF model
internally so this is invisible there. Callers of `from_model` that keep running the HF model afterwards
would be affected. Commits 2 and 3 carry this and can be dropped independently.

## Verification
- `tests/test_llm.py -k "TestQwen3MoeForCausalLM or TestQwen2MoeForCausalLM"` (OPTIMUM_RBLN_TEST_LEVEL=full) on RBLN-CA25: 16 passed on the final head.
- CPU eager check of the HF sparse-MoE block vs the wrapper for qwen3_moe, qwen2_moe, mixtral, qwen3_vl_moe:
  max |diff| ≤ 3.5e-10.
- GPT-OSS: no test or checkpoint cache in the repo; checked on a synthetic module with the same attributes
  (six gate/up buffers equal to the previous views and contiguous, `down_proj` shares HF storage, source released).
- gemma4 received the same mechanical change and was not run (no test in the repo).

## Related Issues
Slack thread on Qwen3-MoE host memory (weight copy vs. transformers 5.x fused layout).

### DIFF
--- a/src/optimum/rbln/transformers/utils/moe.py
+++ b/src/optimum/rbln/transformers/utils/moe.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import torch
-from torch import Tensor
+from torch import Tensor, nn
 
 
 def compute_masked_routing_weight_softmax_first(router_logits: Tensor, top_k: int, renormalize: bool) -> Tensor:
@@ -40,3 +40,15 @@ def compute_masked_routing_weight_topk_first(router_logits: Tensor, top_k: int) 
     masked = torch.zeros_like(router_logits_t, dtype=router_logits.dtype)
     masked.scatter_(0, topk_ids, topk_weights)
     return masked  # [E, T]
+
+
+def split_fused_experts(experts: nn.Module) -> tuple[Tensor, Tensor, Tensor]:
+    # HF packs gate|up along dim 1 of gate_up_proj [E, 2I, H]; custom_moe_glu takes them separately, so the
+    # halves are copied to be contiguous while down_proj [E, H, I] is shared. The fused tensor is released
+    # afterwards: the HF experts module is not run once wrapped, and keeping it would double host memory.
+    gate_up = experts.gate_up_proj.detach()
+    intermediate_dim = gate_up.shape[1] // 2
+    gate = gate_up[:, :intermediate_dim, :].contiguous()
+    up = gate_up[:, intermediate_dim:, :].contiguous()
+    experts.gate_up_proj = None
+    return gate, up, experts.down_proj.detach()
--- a/src/optimum/rbln/transformers/models/gemma4/gemma4_architecture.py
+++ b/src/optimum/rbln/transformers/models/gemma4/gemma4_architecture.py
@@ -549,11 +549,11 @@ class Gemma4Experts(nn.Module):
 
         gate_w_op = gate_w.contiguous()
         up_w_op = up_w.contiguous()
-        down_w_op = down_w.contiguous().clone()
+        down_w_op = down_w.detach()
 
-        self.gate_proj = nn.Linear(self.hidden_size, self.num_experts * self.intermediate_size, bias=False)
-        self.up_proj = nn.Linear(self.hidden_size, self.num_experts * self.intermediate_size, bias=False)
-        self.down_proj = nn.Linear(self.num_experts * self.intermediate_size, self.hidden_size, bias=False)
+        self.gate_proj = nn.Linear(1, 1, bias=False)
+        self.up_proj = nn.Linear(1, 1, bias=False)
+        self.down_proj = nn.Linear(1, 1, bias=False)
         self.gate_proj.weight.data = gate_w_op
         self.up_proj.weight.data = up_w_op
         self.down_proj.weight.data = down_w_op
--- a/src/optimum/rbln/transformers/models/gemma4/gemma4_architecture.py
+++ b/src/optimum/rbln/transformers/models/gemma4/gemma4_architecture.py
@@ -552,12 +552,9 @@ class Gemma4Experts(nn.Module):
         experts.gate_up_proj = None
         down_w_op = down_w.detach()
 
-        self.gate_proj = nn.Linear(1, 1, bias=False)
-        self.up_proj = nn.Linear(1, 1, bias=False)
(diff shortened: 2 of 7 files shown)


src/optimum/rbln/transformers/utils/moe.py
```diff
@@ -40,3 +40,15 @@ def compute_masked_routing_weight_topk_first(router_logits: Tensor, top_k: int)
...
+
+
+def split_fused_experts(experts: nn.Module) -> tuple[Tensor, Tensor, Tensor]:
+    # HF packs gate|up along dim 1 of gate_up_proj [E, 2I, H]; custom_moe_glu takes them separately, so the
+    # halves are copied to be contiguous while down_proj [E, H, I] is shared. The fused tensor is released
+    # afterwards: the HF experts module is not run once wrapped, and keeping it would double host memory.
+    gate_up = experts.gate_up_proj.detach()
+    intermediate_dim = gate_up.shape[1] // 2
+    gate = gate_up[:, :intermediate_dim, :].contiguous()
+    up = gate_up[:, intermediate_dim:, :].contiguous()
```
rebel-jongho, line 52: Just curious : no `.clone()` is needed?

rebel-jongho (approved):
LGTM

src/optimum/rbln/transformers/utils/moe.py
rebel-thkim: Right, no `.clone()` needed here. `gate_up[:, :I, :]` slices dim 1 of `[E, 2I, H]`, so the view is non-contiguous (dim-0 stride is `2I*H` while it holds `I*H` elements per expert). `.contiguous()` on a non-contiguous view allocates a fresh contiguous tensor and copies into it, so `gate`/`up` already own their storage and are not views of `gate_up_proj`; that is what lets us drop `experts.gate_up_proj` right after. Adding `.clone()` on top would make a second identical copy, which is exactly the extra copy dev had and this PR removes.

Checked: `gate_up[:, :I].contiguous().untyped_storage().data_ptr() != gate_up.untyped_storage().data_ptr()` for E > 1. The only case where `.contiguous()` returns the view itself is E == 1 (the slice is already contiguous); then `gate`/`up` keep the fused storage alive, which is still correct, just not freed.

`down_proj` is different: it already has the layout the op wants, so we deliberately share it with the HF module (`.detach()`, no copy).

PR #694, opened 2026-08-26
model: support cosmos-predict2/predict2.5/transfer2.5 pipeline

rebel-seinpark:
# Pull Request Description

> **⚠️ Important: Branch Target**
> - **New features, enhancements, and non-critical fixes**: Merge to `dev` branch
> - **Critical hotfixes only**: Merge to `main` branch (must also merge to `dev`)
> 
> Please ensure you've selected the correct base branch before submitting!

## Type of Change

- [ ] Release (dev → main merge for production release)
- [ ] New Model Support
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe):

## Related Issues

----
# Conventional commit
```
type(optional scope): description
```
----
# Type candidate
  - Model Updates
    - `model`: Adding New models or Bugfix for existing models
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
    - `performance`: Optimizing some models or this library itself
      - ex) Loading RBLNModel faster
      - ex) Optimizing Memory Usage of DecoderOnlyModel
  - Code Refactor
    - `refactor`: Re-arrange class architecture, or more.
      - ex) Refactor Seq2Seq
  - Documentation
    - `doc`: Update docstring only
  - Library Dependencies
    - `dependency`: Update requirements, something like that.
  - Release
    - `release`: Merging dev to main for production release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

PR #715, opened 2026-09-07
model: resolve return_dict=None to config.use_return_dict in raw forwards

rebel-seinpark:
## Problem

Under HF semantics, `return_dict=None` means "follow `config.use_return_dict`" (default True). However, 19 RBLN forwards branched directly on `if not return_dict:`, so **a raw forward call that does not pass `return_dict` returned a plain tuple instead of a ModelOutput**.

- The `generate()` path was unaffected — it passes `return_dict=True` explicitly.
- Pipelines that call a model as a raw forward break. Real case: the Cosmos-Predict2.5 pipeline calls its text encoder (Qwen2.5-VL) as `text_encoder(input_ids, output_hidden_states=True)` → crashes with `'tuple' object has no attribute 'hidden_states'`. (Found during the #694 bring-up.)

This is a follow-up in the same family as #699 (raw-forward defaults for `attention_mask`/`generate_idx`).

## Changes

1. Add the standard HF boilerplate line to every affected forward:
   ```python
   return_dict = return_dict if return_dict is not None else self.config.use_return_dict
   ```
   Affected (12 files, 19 forwards): `decoderonly` (the CausalLM base — applies to every causal LLM), `qwen2_vl`×2, `qwen2_5_vl`×2, `qwen3_vl`×2, `qwen3_5`×2, `exaone4_5`×2, `llava`, `llava_next`, `idefics3`×2, `paligemma`, `blip_2`×2 (vision/qformer), `pixtral`
2. Fix the annotation typo in `RBLNDecoderOnlyModelForCausalLM.forward`: `return_dict: torch.Tensor | None` → `bool | None`
3. `RBLNRuntimePixtralVisionModel.__init__` received `config` but never stored it, so the fallback had no `self.config` to resolve against → added `self.config = config`

## How the targets were selected

An AST scan extracted every forward where the `return_dict` parameter defaults to `None`, the body branches on `if not return_dict:`, and no `use_return_dict` fallback exists. Re-scanning after the fix finds zero remaining instances.

Excluded as false positives:
- The `RBLNModel` base forward, `modeling_generic`, `colpali`, `colqwen2`, `detr`, `grounding_dino` — already have the fallback
- `clip`/`siglip`/`gemma4` — delegate to `RBLNModel.forward`, which has the fallback
- Everything on the diffusers side — `return_dict: bool = True` is diffusers' own convention

## Backward compatibility

The only behavior change: a raw forward called without `return_dict` now returns a ModelOutput instead of a tuple (= the standard HF behavior).

- Audited every internal caller that could depend on tuple returns: only sites passing an explicit `return_dict=False` exist (the t5 runtime wrapper, the siglip compile wrapper), and explicit `False` behaves exactly as before.
- colqwen2/colpali pass `return_dict=True` explicitly — unchanged. The `RBLNDecoderOnlyModel.forward` (L654) that paligemma calls takes no `return_dict` parameter at all — unchanged.

`ruff format` / `ruff check` pass.

### DIFF
--- a/src/optimum/rbln/transformers/models/blip_2/modeling_blip_2.py
+++ b/src/optimum/rbln/transformers/models/blip_2/modeling_blip_2.py
@@ -126,6 +126,7 @@ class RBLNBlip2VisionModel(RBLNModel):
         Returns:
             BaseModelOutputWithPooling or tuple(torch.FloatTensor): The model outputs. If return_dict=False is passed, returns a tuple of tensors. Otherwise, returns a BaseModelOutputWithPooling object.
         """
+        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
         batch_size = pixel_values.shape[0]
         outputs = []
         for i in range(batch_size):
@@ -257,6 +258,7 @@ class RBLNBlip2QFormerModel(RBLNModel):
         Returns:
             BaseModelOutputWithPoolingAndCrossAttentions or tuple(torch.FloatTensor): The model outputs. If `return_dict=False` is passed, returns a tuple of tensors. Otherwise, returns a `BaseModelOutputWithPoolingAndCrossAttentions` object.
         """
+        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
         batch_size = query_embeds.shape[0]
         outputs = []
         for i in range(batch_size):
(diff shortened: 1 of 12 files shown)


rebel-thkim (approved):
👍

PR #722, opened 2026-09-08
model: moe release the safetensors mmap after loading (Qwen3/Qwen2/Mixtral/Qwen3-VL/Gemma4 MoE)

rebel-thkim:
## Type of Change
- [x] Performance improvement

## Changes Overview
Adds `release_checkpoint_mmap_(model)` and `RBLNMoeLoadMixin` (a `get_pytorch_model` classmethod that applies it
after `super().get_pytorch_model`) to `transformers/utils/moe.py`. The mixin is the first base of
`RBLNQwen3MoeForCausalLM`, `RBLNQwen2MoeForCausalLM`, `RBLNMixtralForCausalLM`, `RBLNQwen3VLMoeModel`,
`RBLNQwen3VLMoeForConditionalGeneration`, `RBLNGemma4ForCausalLM` and `RBLNGemma4ForConditionalGeneration`. Right after `from_pretrained` it clones every parameter and buffer
except the fused `experts.gate_up_proj`; models without that tensor (dense) are left untouched.

transformers 5.x loads safetensors through `safe_open(..., backend="mmap")` and installs every weight it does
not convert as a view of that mapping (no copy). For the hub checkpoints of these families the experts are
converted (stacked from per-expert tensors, or transposed for Qwen3-VL-MoE) and become fresh copies, while the
remaining ~3 GiB of views keep every shard mapped, so the 54 GiB of expert pages the conversion read stay charged
to RSS although nothing reads them again. transformers closes the file handles but never unmaps or `madvise`s.

Why skip only `gate_up_proj` rather than the experts module: a checkpoint already stored in the fused layout
(Qwen3-VL-MoE re-saved with `save_pretrained`, or a raw `state_dict()` dump) loads its experts as views too. The
wrapper (#714) copies `gate_up_proj` into gate/up and drops it anyway, so leaving just that tensor and copying the
rest (including `down_proj`) means that once wrapped no view remains in either layout. In the per-expert case the
`down_proj` copy is redundant (18 GiB memcpy on Qwen3-30B-A3B, 0.6 s, no peak increase since each tensor is
replaced in turn).

Buffers need a different treatment from parameters: transformers installs a checkpoint buffer as the mmap view
itself (no `nn.Parameter` wrap), and a view keeps its base alive across a `.data` swap, so buffers are replaced
as objects (`setattr(module, name, buffer.detach().clone())`). Verified on Gemma4-26B-A4B, whose per-layer
`layer_scalar` buffers otherwise kept a 46 GiB shard mapped.

Gemma4 checkpoints are natively fused, so at load everything is a view (nothing to do), but the wrapper's gate/up
copy then leaves the whole shard resident through the remaining views; the same override fixes that. Dense Gemma4
has no `gate_up_proj` and is a no-op. GPT-OSS (own loader) is not covered.

## Motivation and Context
Follow-up to #714. Measured on Qwen/Qwen3-30B-A3B bf16 (checkpoint 56.9 GiB) right after
`from_pretrained`, process-tree RSS in GiB:

| | anon | file (mmap) | total | shard mappings |
|---|---|---|---|---|
| before | 54.4 | 54.1 | 108.5 | 16 |
| after | 57.2 | 0.1 | 57.3 | 0 |

Per family right after `from_pretrained` (bf16, process RSS in GiB). "other" is what gets cloned:

| model | experts (anon) | other (file-backed views) | file RSS before | file RSS after |
|---|---|---|---|---|
| Qwen3-30B-A3B | 54.0 | 2.9 | 54.3 | 0.3 (measured) |
| Qwen1.5-MoE-A2.7B | 23.2 | 3.5 | 23.3 | ~0 (same mechanism) |
| Mixtral-8x7B-Instruct | 84.0 | 3.0 | 84.1 | ~0 (same mechanism) |
| Qwen3-VL-30B-A3B-Instruct | 54.0 | 3.9 | 54.1 | ~0 (same mechanism) |
| gemma-4-26B-A4B-it (natively fused; measured after the wrapper drops `gate_up_proj`) | 0 at load, 28.4 after wrapper | 19.7 (down + others, cloned) | 46.7 | 0.3 (measured) |

Cost: +2.9 GiB anon and one extra read of the non-expert tensors (about 8 s from NAS in the run above;
the compiler would read those pages later anyway). The released 51 GiB was clean page cache, so this
does not change the OOM behaviour; it makes tree_rss / cgroup accounting reflect the memory that is
actually in use, and avoids reclaim churn when the compile runs near a cgroup limit.

## Verification
- `tests/test_llm.py::TestMoeHostMemory` (CPU, collected by CI): builds synthetic 0.4-0.8 GiB checkpoints, then
  asserts that RSS growth through `get_pytorch_model` + `_wrap_model_if_needed` stays under 1.5x the safetensors
  size and that loaded weights equal the source. Covers the per-expert families (Qwen3-MoE, Qwen2-MoE, Mixtral),
  Gemma4's natively fused layout and the Qwen3-VL-MoE hub (transposed) layout. On this branch (with #714 in dev)
  all cases are ~1.0x; on dev before either PR the per-expert cases are ~2.9x and Qwen3-VL-MoE 1.9x, so the test
  fails there.
- Qwen3-30B-A3B bf16 through the export path, process-tree RSS in GiB (dev / #714 only / #714 + #722):
  after `from_pretrained` 108.7 / 109.4 / 58.5, Export done 162.9 / 108.9 / 57.7, MLIR entry 221.6 / 167.6 / 114.8.
- gemma-4-26B-A4B-it through `RBLNGemma4ForConditionalGeneration.get_pytorch_model`: anon 20.1 / file 18.4 with
  only the 30 `gate_up_proj` views left; dropping them (what the wrapper does) leaves file 0.3 and 0 mappings.
  gemma-4-31B-it (dense): no-op.
- `TestQwen3MoeForCausalLM` / `TestQwen2MoeForCausalLM` on RBLN-CA25 passed on earlier heads; CI runs the LLM job on
  the final head.

## Related Issues
Follow-up to #714. Slack thread on Qwen3-MoE host memory (weight copy vs. transformers 5.x fused layout).


tests/test_release_checkpoint_mmap.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_fused_checkpoint_loads_equal(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            src = Qwen3MoeForCausalLM(CONFIG).eval()
+            save_file({k: v.contiguous() for k, v in src.state_dict().items()}, f"{tmp}/model.safetensors")
+            src.config.save_pretrained(tmp)
+            self.assertIn("model.layers.0.mlp.experts.gate_up_proj", _safetensors_keys(tmp))
+
+            model = RBLNQwen3MoeForCausalLM.get_pytorch_model(tmp)
+            for (name, p), (_, q) in zip(model.named_parameters(), src.named_parameters(), strict=True):
```
rebel-jongho: 1. is this test file covered by ci? 

2. do we need this?

src/optimum/rbln/transformers/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -66,3 +67,7 @@ class RBLNQwen3MoeForCausalLM(RBLNDecoderOnlyModelForCausalLM):
     """
 
     _decoder_wrapper_cls = Qwen3MoeWrapper
+
+    @classmethod
+    def get_pytorch_model(cls, *args, **kwargs):
+        return release_checkpoint_mmap_(super().get_pytorch_model(*args, **kwargs), "Qwen3MoeExperts")
```
rebel-jongho: Is needed only for moe models? why not other llm models?

tests/test_release_checkpoint_mmap.py
rebel-thkim: 1. It was not: the CI workflow runs `tests/test_config.py`, `test_transformers.py`, `test_diffusers.py` and `test_llm.py` by name (`rbln_optimum_pytest.yaml:173-187`), so a new file would have been skipped. Moved the test into `tests/test_llm.py` as `TestReleaseCheckpointMmap` (f1471de0). It is CPU-only and builds tiny synthetic Qwen3-MoE / Qwen2-MoE / Mixtral checkpoints, so it adds a few seconds to the LLM job.

2. The fused-layout case is dropped. It only covered a non-standard checkpoint (a raw `state_dict()` dump; `save_pretrained` always writes the per-expert layout), and for that layout the clone has no effect to test beyond "values are unchanged". What remains is the case the PR is for: a per-expert checkpoint loads with no tensor left viewing the safetensors mmap, and the weights equal the source model.

src/optimum/rbln/transformers/models/qwen3_moe/modeling_qwen3_moe.py
rebel-thkim: Only for models whose experts get converted at load. transformers 5.x installs every weight it does not convert as a view of the safetensors mmap (no copy). For a dense model that is *all* weights, so the mmap is the model: cloning would turn N GiB of reclaimable page cache into N GiB of anonymous memory with no change in RSS, and it would be worse under memory pressure because anon cannot be dropped. Measured on DeepSeek-R1-Distill-Qwen-32B (dense, bf16): right after `from_pretrained` anon 0.4 GiB, 0% of the checkpoint pages resident, 100% of parameters file-backed. Nothing to release.

The MoE families are different because transformers rebuilds the experts (stack of per-expert tensors, or transpose for Qwen3-VL-MoE) into new memory. That read faults the whole checkpoint into RSS, and afterwards only the small non-expert tensors (2.9 GiB on Qwen3-30B-A3B) still view the mmap, yet they keep all 54 GiB mapped. Cloning those few tensors is what unmaps it. Gemma4 and GPT-OSS are excluded for the same reason as dense models: their experts load as views (fused-layout checkpoint / own loader), so there is nothing dead to drop.

If dense models need the mmap gone for accounting reasons, the right tool is `madvise`/`posix_fadvise` on the mapping rather than cloning, and that would be a separate change.

rebel-jongho: Do all of moe familes in this PR have same behavior? not checkpoint-specific?

+ madvise/posix is not a right choice.

rebel-thkim: It is checkpoint-layout dependent, not a property of the model class, so let me be precise about what was checked.

The clone is only useful when transformers converted the experts at load (so they are already fresh copies) and the remaining weights are mmap views. Verified on the hub checkpoints of all four families (right after `from_pretrained`, per tensor via `/proc/self/maps`):

| checkpoint | experts | everything else |
|---|---|---|
| Qwen/Qwen3-30B-A3B | 100% anon (stacked) | 100% file-backed, 2.9 GiB |
| Qwen/Qwen1.5-MoE-A2.7B | 100% anon (stacked) | 100% file-backed, 3.5 GiB |
| mistralai/Mixtral-8x7B-Instruct-v0.1 | 100% anon (stacked) | 100% file-backed, 3.0 GiB |
| Qwen/Qwen3-VL-30B-A3B-Instruct | 100% anon (transposed) | 100% file-backed, 3.9 GiB |

Checkpoints re-saved with TF5 `save_pretrained`: Qwen3-MoE / Qwen2-MoE / Mixtral are written back per-expert (`SplitModulelist` reverse op), so they load the same way; the unit test builds its fixtures exactly like that. Qwen3-VL-MoE is the exception: `Transpose(check_dims=True)` is shape-driven, so `save_pretrained` writes the fused `[E, 2I, H]` layout as is, and reloading it installs the experts as mmap views too. In that case the clone copies the ~3 GiB of non-expert weights and releases nothing (the mapping stays alive through the expert views); values are unchanged, so it is harmless but not useful. Same for any fused-layout checkpoint (raw `state_dict()` dump). I also found while checking this that the override was missing on `RBLNQwen3VLMoeForConditionalGeneration` (only `RBLNQwen3VLMoeModel` had it); fixed in 4ed194ce with a test that builds the hub (transposed) layout.

On madvise/posix_fadvise: agreed, I take that back. For a dense model the mmap pages are the live weights until the compiler makes its own copy, so dropping them would only force re-reads, and `posix_fadvise` does nothing for mapped pages. Nothing on the optimum side to release there; the file RSS that remains after the compiler copy is the compiler's copy to avoid (borrow / weight-free), not ours to drop.

rebel-thkim: In #714, since the custom op creates an "expert clone" and breaks the reference to the original in order to achieve the desired layout, the result is the creation of two sets—file-rss and anon-rss—which is a common behavior of the moe model, regardless of the checkpoint. Regarding madvise/posix, Claude provided an answer without considering my intentions, but I expect that file-rss will be sufficient for dense models.

rebel-jongho (commented):
How about use `Mixin` pattern for all MoE classes, instead of wrapping `get_pytorch_model` of every class? 

and also Please check that ci is failing

tests/test_llm.py
```diff
@@ -1288,5 +1313,142 @@ class TestDisallowedLlama_4(DisallowedTestBase.DisallowedTest):
     RBLN_CLASS_KWARGS = {"rbln_config": {"attn_impl": "flash_attn", "kvcache_partition_len": 2048}}
 
 
+@unittest.skipUnless(sys.platform == "linux", "reads /proc/self/status")
```
rebel-jongho: please remove this line so that we can recognize if this test goes wrong

rebel-jongho (approved):
LGTM

Issue #720, opened 2026-09-08
Default (eager) attention with max_seq_len=32768 passes validation but crashes the native compiler on RBLN-CR13


rebel-junamsong:
## The generic eager bound admits a config RBLN-CR13 can never compile

`validate_attention_method` caps eager attention at `DEFAULT_MAX_EAGER_ATTN_SEQUENCE_LENGTH = 32_768` with a strict `>` (`modeling_attention_utils.py:79`), so `max_seq_len=32768` passes validation. But on RBLN-CR13+ the eager attention lowering indexes the sequence axis with int16 (rtosa `nn_matmul` requires `dynamic_axis_extent <= 32767`), so the boundary case sails through Python validation and dies inside the native compiler with no Python traceback:

```
Process ... died with <Signals.SIGABRT: 6>
```

On CA22 the same config compiles and runs (the ATOM twins of the failing suites pass eager-32768), so the effective eager ceiling is device-tiered: **32768 on CA, 32767 on CR13+**.

Since `attn_impl` defaults to `eager`, every model whose default export asks for a 32k context hits this with no user mistake to point at. Observed at scale: the REBEL exaone-3.5-2.4b / qwen2.5 / qwen3 32k suites in rbln_deploy nightly #1443 all crashed exactly here (SIGABRT during prefill compile). functionality-CI never sees it because its CR13 32k cases all pin `-kv 16384` (flash) — the eager-32k *configuration* has no CI coverage anywhere.

## Suggested fix (two sites)

1. **`set_default_values`** — when `attn_impl` is unset, `max_seq_len >= 32768`, and the target is CR13+ (`npu_is_cr13_or_later`), default to `flash_attn` instead of eager: the same auto-switch the function already performs when `kvcache_partition_len` is provided (`modeling_attention_utils.py:46-53`). The upgraded default (`flash_attn` + partition 16384) is exactly the configuration verified on CR13 hardware in deploy manual run #1452.
2. **`validate_attention_method`** — an explicit `attn_impl="eager"` with `max_seq_len >= 32768` on CR13+ should raise the existing descriptive `ValueError` (with the real 32767 ceiling) instead of deferring to the native crash.

## Notes

[...]

PR #729, opened 2026-09-09
model(decoderonly): per-family attention limits (ATOM 32k / REBEL 16k)

rebel-junamsong:
## Issue

One set of globals bounded every target — `DEFAULT_MAX_EAGER_ATTN_SEQUENCE_LENGTH`,
`MAX_FLASH_ATTN_PARTITION_LENGTH`, `MIN_FLASH_ATTN_*`, `DEFAULT_FLASH_ATTN_PARTITION_LENGTH`,
`MAX_SLIDING_WINDOW_SIZE`.

They match ATOM. On REBEL they are a tier too wide, so the config passes Python validation and the
native compiler aborts — `exit 134`, core dump, no Python traceback. deploy nightly #1443: 15 REBEL
32k suites died there while the ATOM twins compiled.

The bounded quantity is the extent of the KV cache's dynamic axis, and each attention mode makes a
different parameter that axis: `max_seq_len` (eager), `kvcache_partition_len` (flash),
`sliding_window` (sliding-window cache). `validate_attention_method` had no `npu` to see the family
with.

Refs #720. Supersedes #721 (closed).

## Resolution

`AttentionLimits` — one frozen record per family, resolved from the target NPU:

| bounded parameter | ATOM (`RBLN-CA`) | REBEL (`RBLN-CR`) |
|---|---|---|
| eager `max_seq_len` | ≤ 32,768 | ≤ 16,384 |
| flash `kvcache_partition_len` | 1,024 – 32,768 | 1,024 – 16,384 |
| flash default partition | 16,384 | 8,192 |
| `sliding_window` (− prefill chunk) | ≤ 32,640 | ≤ 32,255 |

- `get_attention_limits(npu)`: `RBLN-CR*` → REBEL, `RBLN-CA*` → ATOM, no NPU → ATOM (the compiler
  stays the backstop), unknown name → `ValueError` rather than the wider limits.
- `validate_attention_method` gains an optional `npu`, wired from its single caller; every message
  names the resolved family and value.
- `min_flash_max_seq_len` is derived (`2 × min_flash_partition_len`), and one
  `resolve_npu_or_none` in `runtime_utils` replaces three drifted copies of the npu resolution.

ATOM behaviour is unchanged throughout.

**Why REBEL 16,384 and not the measured 32,767 ceiling** — both bounded parameters are powers of two
in practice (`kvcache_partition_len` must divide `max_seq_len`; eager's bound *is* `max_seq_len`), so
16,384 is the largest usable value under it. Tradeoff: eager `max_seq_len` in (16,384, 32,767]
compiles on REBEL today and is now rejected.

**Why the default halves to 8,192** — at 16,384 a REBEL default flash config sat exactly on the cap,
so the only valid `max_seq_len` was 32,768 or more.

**Why `sliding_window` keeps the hard ceiling (32,767) instead of 16,384** — the grid argument does
not reach that axis. `sliding_window` comes from the model config and has to divide nothing, so the
INT16 ceiling is its only bound; `max_sliding_window` is 32,768 on ATOM and 32,767 on REBEL, exactly
the two measured ceilings, and the REBEL bound moves by one token rather than a tier.

## Verification

- `TestAttentionLimits`, 6 functions — one per contract clause: family resolution (both families,
  unknown-npu raise, attached-device fallback, pinned target wins), the eager cap, the flash
  partition range, the per-family default, the record's internal consistency, the sliding-window bound.
- `test_attention_limits_npu_wiring`: a pinned REBEL target rejects eager 32k before
  `get_compiled_model` runs, so `rbln_config.npu` provably reaches the validator without compiling.
- Local: every assertion above run against the real module; `ruff format --check .` and
  `ruff check .` (0.15.4, as locked) clean.

## Follow-up

Auto-promoting an unset `attn_impl` to `flash_attn` above the eager cap went away with #721. It is
worth its own PR, with the trigger read from `limits.max_eager_seq_len` rather than a hardcoded
32,768; branch `fix/cr13-eager-32k-flash-default` is kept for it.

### DIFF
--- a/src/optimum/rbln/transformers/modeling_attention_utils.py
+++ b/src/optimum/rbln/transformers/modeling_attention_utils.py
@@ -17,7 +17,6 @@ if TYPE_CHECKING:
 logger = get_logger()
 
 
-DEFAULT_FLASH_ATTN_PARTITION_LENGTH = 16_384
 MAX_SLIDING_WINDOW_SIZE = 32_768
 
 
@@ -34,6 +33,7 @@ class AttentionLimits:
     max_eager_seq_len: int
     min_flash_partition_len: int
     max_flash_partition_len: int
+    default_flash_partition_len: int
 
     @property
     def min_flash_max_seq_len(self) -> int:
@@ -46,6 +46,7 @@ ATOM_ATTENTION_LIMITS = AttentionLimits(
     max_eager_seq_len=32_768,
     min_flash_partition_len=1_024,
     max_flash_partition_len=32_768,
+    default_flash_partition_len=16_384,
 )
 
 REBEL_ATTENTION_LIMITS = AttentionLimits(
@@ -53,6 +54,7 @@ REBEL_ATTENTION_LIMITS = AttentionLimits(
     max_eager_seq_len=16_384,
     min_flash_partition_len=1_024,
     max_flash_partition_len=16_384,
+    default_flash_partition_len=8_192,
 )
 
 
@@ -98,7 +100,7 @@ def set_default_values(
             )
 
     if kvcache_partition_len is None and attn_impl == "flash_attn":
-        kvcache_partition_len = DEFAULT_FLASH_ATTN_PARTITION_LENGTH
+        kvcache_partition_len = get_attention_limits(npu).default_flash_partition_len
 
     if kvcache_block_size is None:
         if attn_impl == "eager":
--- a/src/optimum/rbln/transformers/modeling_attention_utils.py
+++ b/src/optimum/rbln/transformers/modeling_attention_utils.py
@@ -54,7 +54,7 @@ REBEL_ATTENTION_LIMITS = AttentionLimits(
     min_flash_partition_len=1_024,
     max_flash_partition_len=16_384,
     default_flash_partition_len=8_192,
-    max_sliding_window=16_384,
+    max_sliding_window=32_767,
 )
 
 
--- a/src/optimum/rbln/transformers/modeling_attention_utils.py
+++ b/src/optimum/rbln/transformers/modeling_attention_utils.py
@@ -1,6 +1,7 @@
 import logging
 import math
 from collections import defaultdict
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import rebel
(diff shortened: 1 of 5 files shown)


src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -16,14 +17,60 @@
...
+
+def get_attention_limits(npu: str | None = None) -> AttentionLimits:
+    """Attention limits of the target NPU, falling back to the attached device.
+
+    ATOM accepts twice the dynamic-axis extent REBEL does, so the two families carry separate
+    limits. When the family cannot be resolved — compiling on a host without an NPU and without
+    `npu` pinned on the config — the wider ATOM limits apply and the compiler stays the backstop.
+    """
+    npu = npu or (rebel.get_npu_name(0) if rebel.npu_is_available(0) else None)
+    if npu is not None and npu.startswith("RBLN-CR"):
```
rebel-junamsong: An unrecognised `npu` silently gets ATOM's wider limits. `runtime_utils._dram_spec` answers the same question and refuses:

```python
if npu.startswith("RBLN-CR"): ...
elif npu.startswith("RBLN-CA"): ...
raise ValueError(f"Unknown npu name: {npu}")
```

The docstring justifies the fallback for *no device attached*, which is `npu is None` and fine. A typo'd or future `npu` pinned on the config is a different case and lands on the wider limits — so validation passes and the compiler aborts instead, which is the failure this PR exists to prevent (deploy #1443).

```python
if npu is None:
    return ATOM_ATTENTION_LIMITS  # no device; compiler is the backstop
if npu.startswith("RBLN-CR"):
    return REBEL_ATTENTION_LIMITS
if npu.startswith("RBLN-CA"):
    return ATOM_ATTENTION_LIMITS
raise ValueError(f"Unknown npu name: {npu}")
```

src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -16,14 +17,60 @@
...
+
+
+def get_attention_limits(npu: str | None = None) -> AttentionLimits:
+    """Attention limits of the target NPU, falling back to the attached device.
+
+    ATOM accepts twice the dynamic-axis extent REBEL does, so the two families carry separate
+    limits. When the family cannot be resolved — compiling on a host without an NPU and without
+    `npu` pinned on the config — the wider ATOM limits apply and the compiler stays the backstop.
+    """
+    npu = npu or (rebel.get_npu_name(0) if rebel.npu_is_available(0) else None)
```
rebel-junamsong: Third copy of this resolution, and the three disagree:

- here — guarded, falls back to `None`
- `runtime_utils.npu_is_cr13_or_later` — this line verbatim
- `set_default_values` L88 — `npu = npu or rebel.get_npu_name(0)`, **unguarded**, and only when `prefill_chunk_size is None`

So `get_attention_limits(npu)` twelve lines later reads an `npu` that was resolved or not depending on an unrelated argument, and on a host with no NPU L88 raises while this line returns `None`. `runtime_utils._resolve_npu` already exists; one shared helper called once at the top of `set_default_values` would remove the ordering question entirely.

src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -64,23 +111,34 @@ def set_default_values(
...
     if attn_impl not in ["eager", "flash_attn"]:
         raise ValueError(f"Unknown `attn_impl` : {attn_impl}. (Available : 'eager', 'flash_attn`)")
 
+    limits = get_attention_limits(npu)
+
     ## Checking Constraints...
+    # The device bounds the extent of the KV cache's dynamic axis, which eager attention sets to
+    # `max_seq_len` and flash attention to `kvcache_partition_len`. ATOM allows 32k, REBEL 16k.
+
     # Constraint of eager attention:
```
rebel-junamsong: These comments used to carry the numbers (`1k <= partition_len <= 32k`, `at least 2048`); now they restate the identifiers that appear in the `if` six lines down, so they say nothing the code doesn't. The line above already gives the fact worth keeping — "ATOM allows 32k, REBEL 16k". Suggest dropping the symbolic restatement and leaving that one.

Also `## Checking Constraints...` is a stray double-hash, and the signature you just touched still has no `-> None`.

src/optimum/rbln/transformers/models/decoderonly/configuration_decoderonly.py
```diff
@@ -159,11 +160,12 @@ def __init__(
             `kvcache_partition_len` is relevant **only** when `attn_impl` is `"flash_attn"`.
 
             - It defines the length (number of tokens) of each partition within the Key-Value (KV) cache.
-            - Must be between 1,024 and 32,768 (inclusive).
+            - Must be between 1,024 and the target NPU's limit, inclusive — 32,768 on ATOM
```
rebel-junamsong: The per-family numbers are now written out in three places here (L146-147, L163-164, L168) — the drift the `AttentionLimits` dataclass was introduced to prevent, reintroduced in prose. Nothing fails if `REBEL_ATTENTION_LIMITS` changes and these don't.

Cheapest fix is to stop repeating the pairs and point at the source once ("see `AttentionLimits` in `modeling_attention_utils`"), keeping only the shape of the rule in prose. Related: L150 still hardcodes "at least 2,048", which is now `2 * min_flash_partition_len` — correct today only because both families share the 1,024 minimum.

rebel-junamsong (commented):
Frozen dataclass + a derived `min_flash_max_seq_len` property is the right shape here — it kills the implicit `2048 = 2 × 1024` coupling the old constant pair left unstated, and `str | None` throughout is current. Tests cover both caps, both defaults, and all three resolution paths.

Four things, one of them behavioural. Formatting nit not worth its own thread: the source writes `32_768` / `1_024` but the new tests write `32768` / `1024` — worth matching within the PR.

src/optimum/rbln/transformers/modeling_attention_utils.py
```diff
@@ -16,12 +17,63 @@
...
+    max_sliding_window=32_768,
+)
+
+REBEL_ATTENTION_LIMITS = AttentionLimits(
+    name="REBEL",
+    max_eager_seq_len=16_384,
+    min_flash_partition_len=1_024,
+    max_flash_partition_len=16_384,
+    default_flash_partition_len=8_192,
+    max_sliding_window=16_384,
```
rebel-thkim: should be 32768?

rebel-thkim: 32767?

rebel-thkim:
```suggestion
    max_sliding_window=32_767,
```

rebel-thkim (approved):
👍

PR #734, opened 2026-09-10
refactor: register public objects through per-module __all__ and define_import_structure

rebel-jongho:
# Pull Request Description

## Type of Change
- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

## Changes Overview

Every package `__init__.py` under `optimum.rbln` now follows the layout transformers uses since v4.48: the module that defines a public object lists it in `__all__`, and the package `__init__.py` is a fixed template.

```python
if TYPE_CHECKING:
    from .configuration_llama import *
    from .modeling_llama import *
else:
    import sys

    _file = globals()["__file__"]
    sys.modules[__name__] = _LazyModule(__name__, _file, define_import_structure(_file), module_spec=__spec__)
```

The hand-maintained `_import_structure` dictionaries at the top level, in `transformers`, `transformers/models`, `diffusers` and the diffusers subpackages are gone, and the eagerly importing leaf `__init__.py` files are lazy now too. Adding a model no longer touches four `__init__.py` files: write `__all__` in the new modules and run `python scripts/check_inits.py --fix`.

- `src/optimum/rbln/utils/import_utils.py` gains `define_import_structure`. It reuses `transformers.utils.import_utils.create_import_structure_from_path` to discover the `__all__` entries, but flattens the nested result itself. transformers' own `spread_import_structure` silently drops modules when a package holds both modules and subpackages at the same level. In this tree it lost every `configuration_*` module under `transformers/models` and `diffusers.pipelines.auto_pipeline`.
- `scripts/check_inits.py` verifies that the `TYPE_CHECKING` block of every package matches the children that export something, that every `__all__` entry is defined in its module, that `__all__` is written in the one-name-per-line form transformers' parser accepts, and that a package which exports objects uses the lazy template. `--fix` rewrites the blocks. Both quality gates (`check_code_quality.yml` and the `:mag: ruff` step of `.buildkite/pr.yml`) run it.
- The `__init__.py` files shrink from 3,225 lines to 2,316 lines. The top-level one goes from 623 lines to 45.

## Motivation and Context

Registering a public class required editing the model package `__init__.py`, `transformers/models/__init__.py`, `transformers/__init__.py` and the top-level `__init__.py`, each holding both a string list and a `TYPE_CHECKING` import block. The lists drifted: `RBLNQwen3_5TextModel` and `RBLNQwen3_5TextModelConfig` were registered in `qwen3_5/__init__.py` but never reached `optimum.rbln`. A single `__all__` next to the class definition removes the duplication, and the checker catches the remaining piece (the type-checking block) in CI.

## Verification

The public surface was dumped on `dev` and on this branch for `optimum.rbln`, `optimum.rbln.transformers`, `optimum.rbln.diffusers` and all 74 subpackages: every public name in `dir()` together with the `__module__` of the object it resolves to. The two dumps are identical except for these intended differences.

| Package | Difference |
| --- | --- |
| `optimum.rbln`, `optimum.rbln.transformers`, `optimum.rbln.transformers.models` | `RBLNQwen3_5TextModel` and `RBLNQwen3_5TextModelConfig` are now exported, matching the other Qwen3.5 classes. |
| `optimum.rbln` | `modeling_rope_utils` no longer shows up in `dir()` before it is imported. |
| `...models.exaone`, `...models.midm` | `os`, `environ`, `this_path`, `local_dir` are gone. These inits set `LOCAL_CACHE_ROOT_CUSTOM_CODE_MIDM` to a directory that does not exist and nothing reads the variable. |

As in transformers, `dir()` of a lazy package additionally lists the dotted module paths it can load (for example `transformers.models.llama.modeling_llama`).

- `from optimum.rbln import *`, `optimum-rbln-cli --list-classes` (149 classes, `RBLNQwen3_5TextModel` now among them) and `--class RBLNQwen3ForCausalLM --show-rbln-config` work.
- `python -X importtime -c "import optimum.rbln"`: 1.71 s on `dev`, 1.75 s here, both dominated by `rebel`. The number of modules loaded is the same (3,123).
- `pytest tests/test_base.py tests/test_config.py`: 58 passed, 4 skipped, 9 failed. The 9 failures are the `test_deprecate_method_raises_at_or_past_cutoff` cases, which fail identically on `dev` (`unittest.mock.patch` cannot resolve `optimum.rbln.utils` through `_LazyModule`); `dev` additionally failed 3 `test_config` cases that need an NPU.
- `.buildkite/scripts/check-docstrings.sh` passes on every changed `src` file.
- `ruff format --check .`, `ruff check .` and `python scripts/check_inits.py` pass.

## Related Issues


src/optimum/rbln/diffusers/configurations/pipelines/__init__.py
```diff
@@ -1,34 +1,34 @@
...
-)
-from .configuration_stable_diffusion_xl import (
-    RBLNStableDiffusionXLImg2ImgPipelineConfig,
-    RBLNStableDiffusionXLInpaintPipelineConfig,
-    RBLNStableDiffusionXLPipelineConfig,
-)
-from .configuration_stable_video_diffusion import (
-    RBLNStableVideoDiffusionPipelineConfig,
-)
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-thkim: 2026?

rebel-jongho: Right, these headers are new in this PR (the file had none before), so they should read 2026. Fixed in the latest commit for all eight files that gained a header here, including scripts/check_inits.py.

rebel-thkim (approved):
👍

Issue #684, opened 2026-08-24
[MODEL] Qwen3-ASR-1.7B


rebel-hjyang:
**Note**: I have already ported this model to ATOM-Max — full working code (optimum-rbln port + vllm-rbln integration, compile → vLLM serve validated end-to-end) is available at https://github.com/rebellions-sw/Qwen3-ASR-1.7B-FDE.

### Model Information
- Model name: Qwen3-ASR-1.7B
- Model link (HuggingFace/Paper/GitHub): https://huggingface.co/Qwen/Qwen3-ASR-1.7B-hf (transformers-native layout; original: https://huggingface.co/Qwen/Qwen3-ASR-1.7B)
- Task type: Automatic Speech Recognition (multilingual speech-to-text, 30 languages, auto language detection)

### Model Details
- Input modality (Text/Image/Audio/Other): Audio (16 kHz log-mel spectrogram) + Text (prompt/context)
- Output modality (Text/Image/Audio/Other): Text (`language <NAME><asr_text>` + transcription)
- Model size (parameters): ~2.0B total = audio encoder ~0.3B (24-layer Whisper-style, windowed attention) + MM projector + Qwen3-1.7B text decoder
- Required memory (if known): ~3.6 GB weights (bf16); measured 4.8 GiB resident on one ATOM-Max including KV cache (batch 1, max_seq_len 8192)

### Input/Output Tensor Information
- Can input/output shapes be inferred from model config? Yes
- If yes, please explain the relationship:
  ```
  Audio encoder:
  - input_features: (batch, num_mel_bins, mel_len)
    - num_mel_bins from audio_config.num_mel_bins (=128)
    - mel_len padded to a multiple of 2 * audio_config.n_window (=100); 10 ms hop -> 100 frames/s
  - input_features_mask: (batch, mel_len)
  - encoder output: (num_audio_tokens, audio_config.output_dim (=2048))
    - num_audio_tokens = 13 tokens per 100 valid mel frames (i.e., 13 tokens/s of audio)
    - attention is windowed: block size = 104 tokens, no cross-window attention

  Text decoder (standard Qwen3, sequential RoPE — no mrope):
  - inputs_embeds: (batch, seq_len, text_config.hidden_size (=2048))
    - audio embeddings scattered at <|audio_pad|> placeholder positions (audio_token_id=151676)
  - logits: (batch, seq_len, text_config.vocab_size (=151936))

[...]

PR #735, opened 2026-09-10
model: enable Qwen3-ASR

rebel-dkhong:
# Pull Request Description

> 

## Type of Change

- [x] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Changes Overview

Enable Qwen3-AST
- Reuse `RBLNQwen3ForCausalLM`
- Implement `RBLNQwen3ASREncoder`

## Related Issues

https://github.com/RBLN-SW/optimum-rbln/issues/684

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update


src/optimum/rbln/transformers/models/qwen3_asr/configuration_qwen3_asr.py
```diff
@@ -0,0 +1,77 @@
...
+        Args:
+            use_inputs_embeds (bool): Whether or not to use `inputs_embeds` as input. Defaults to `True`.
+            audio_tower (RBLNModelConfig | None): Configuration for the audio encoder component.
+            kwargs: Additional arguments passed to the parent `RBLNQwen3ForCausalLMConfig`.
+
+        Raises:
+            ValueError: If `use_inputs_embeds` is False.
+        """
+        super().__init__(use_inputs_embeds=use_inputs_embeds, **kwargs)
+        if not self.use_inputs_embeds:
```
rebel-thkim, line 72: If I'm going to use `self.use_inputs_embeds`, don't I need to explicitly specify `use_inputs_embeds` in `kwargs`?

src/optimum/rbln/transformers/models/qwen3_asr/modeling_qwen3_asr.py
```diff
@@ -0,0 +1,416 @@
...
+    def _create_embedding_layer(self):
+        text_config = self.config.get_text_config()
+        with no_init_weights():
+            embed_tokens = nn.Embedding(
+                text_config.vocab_size,
+                text_config.hidden_size,
+                text_config.pad_token_id,
+                dtype=self.rbln_config.dtype,
+            )
+        return embed_tokens
```
rebel-thkim, line 241: .to(rbln_config.dtype) ??  I think there was a PR that required something like this? @rebel-eunji

rebel-dkhong: Is it https://github.com/RBLN-SW/optimum-rbln/pull/674 ?
According to this PR, we can use `dtype=self.rbln_config.dtype` when using `nn.Embedding`, while we should use `.to(self.rbln_config.dtype)` when using custom embedding classes such as `Gemma3TextScaledWordEmbedding`

src/optimum/rbln/transformers/models/qwen3_asr/configuration_qwen3_asr.py
rebel-dkhong: With the above implementation, `use_inputs_embeds` is initialized in the __init__ method of `RBLNDecoderOnlyModelConfig` and we can use the instance variable like that.

rebel-dkhong: I'm sorry, I misunderstood.
The super class already takes `use_inputs_embeds`, but its default resolves to False.
This model requires `use_inputs_embeds=True`, since the audio embeddings are merged into the text embeddings before the decoder runs.
Without declaring it here, every export would hit the ValueError unless the user explicitly passed `use_inputs_embeds=True` in rbln_config.
So, the class gets `use_inputs_embeds` as an argument to set its default value as True.

src/optimum/rbln/transformers/models/qwen3_asr/qwen3_asr_architecture.py
```diff
@@ -0,0 +1,137 @@
...
+
+        hidden_states = conv_out.view(num_windows, -1, conv_out.shape[-1])
+
+        attn_bias = (1.0 - attn_mask) * torch.finfo(hidden_states.dtype).min
+        for layer in self.layers:
+            hidden_states = layer(hidden_states, attn_bias)
+        return self.multi_modal_projector(self.ln_post(hidden_states))
+
+
+# The base looks for `model.model.layers`; here the decoder is nested one level deeper.
```
rebel-seinpark, line 131: Please minimize unnecessary comments that arise during the development.

src/optimum/rbln/transformers/models/qwen3_asr/configuration_qwen3_asr.py
rebel-seinpark: How about set `super().__init__(use_inputs_embeds=True, **kwargs)`? In cases like this, wouldn't it be better to require users to do something rather than giving them a choice?

rebel-dkhong: In that case, users may don't know why it works when they set `use_inputs_embeds=False` explicitly.
I think it would be better to set `use_inputs_embeds=True` by default and raise error when they set the variable with wrong value.

src/optimum/rbln/transformers/models/qwen3_asr/modeling_qwen3_asr.py
```diff
@@ -0,0 +1,409 @@
...
+    Converts log-mel audio features into audio embeddings in the language model's embedding
+    space.
+    """
+
+    auto_model_class = None
+
+    def __post_init__(self, **kwargs: Any):
+        self.window_size = get_window_size(self.config)
+        self.chunk_len = self.config.n_window * 2
+        self.chunks_per_window = self.config.n_window_infer // self.chunk_len
```
rebel-jongho: please use property + cache(if needed) instead to prevent inconsist values

src/optimum/rbln/transformers/models/qwen3_asr/modeling_qwen3_asr.py
```diff
@@ -0,0 +1,409 @@
...
+    def _wrap_model_if_needed(cls, model: "PreTrainedModel", rbln_config: RBLNQwen3ASREncoderConfig):
+        return Qwen3ASREncoderWrapper(model, rbln_config).eval()
+
+    @classmethod
+    def _update_rbln_config(
+        cls,
+        preprocessors: Union["AutoFeatureExtractor", "AutoProcessor", "AutoTokenizer"] | None = None,
+        model: Optional["PreTrainedModel"] = None,
+        model_config: Optional["PretrainedConfig"] = None,
+        rbln_config: RBLNQwen3ASREncoderConfig | None = None,
```
rebel-jongho: please use type hint in consistently ( No `Optional` )

src/optimum/rbln/transformers/models/qwen3_asr/modeling_qwen3_asr.py
rebel-dkhong: Please review the update that applies your feedback.

rebel-jongho (approved):
lgtm

PR #737, opened 2026-09-11
other: ci - move PR, nightly and BC testing to Obedients and delete the GHA workflows they replace

rebel-thkim:
## Summary
Finishes the Obedients move: #738 and #739 merged into this branch, and the GHA workflows they replace are deleted here.

**Deleted** — `rbln_trigger_on_pr.yaml`, `rbln_scheduled_test.yaml`, `bc_full_inference.yaml`, `bc_inference.yaml`, `bc_latest_tag.yaml`, `rbln_optimum_full_test.yaml`, `slack_report.yaml`, `check_code_quality.yml`, `test-docstrings.yml`, and `parsing_latest_version.py` (bc_latest_tag was its only user).

**Kept** — what talks to GitHub or ships a release: `deploy`, `deploy-on-tag` and the `bc_compile` it calls on a tag, the compiler-update and dependency bots, and the PR title check. `bc_compile` still needs `rbln_optimum_pytest.yaml` and the Harbor CI image, so both stay with it.

### Phase 3a — full test level

- `.buildkite/pipelines/pytest/pipeline.yml` takes `OPTIMUM_RBLN_TEST_LEVEL` from the uploading job (`${OPTIMUM_RBLN_TEST_LEVEL:-default}`), so one file serves both levels. `build.env()` cannot see a pipeline `env:` block, so the value is baked in at upload time and each step's `if` is already a constant when Buildkite evaluates it.
- `.buildkite/pr.yml`: a `[pytest-full]` commit exports `OPTIMUM_RBLN_TEST_LEVEL=full` and uploads **without** `--git-diff-base`, so nothing is path-gated away. Every other PR is unchanged.
- Two new suites, `cli-argument-parsing` and `cli-error-handling`, carry an `if` that is a constant by the time Buildkite evaluates it, so they exist only in a full build. GHA spawns them at every level and skips them inside the job, which holds an NPU just to skip.
- `uv sync` no longer forces an editable rebuild. Reusing another pod's build is what we want (the install points at this checkout), and the metadata fallback already covers the generated `__version__.py`. That was ~18s of the ~56s each suite spent before its first test.

A full build is therefore 10 NPU steps instead of 8.

## Not in this PR
- HF hub tests (`test_push_to_hub`) stay skipped: `HF_USER_ID` / `HF_AUTH_TOKEN` have no Obedients vault entries yet, so they are not mapped. GHA passes `MODEL_HUB_ID` / `MODEL_HUB_ACCESS` here. Once the vault entries exist, adding two lines to the secrets anchor turns them on.
- The GHA `optimum-rbln-full-pytest` job stays until the nightly moves over, so a `[pytest-full]` PR runs both for now.

## Careful
Writing the full-test directive into a commit message makes that build a full one — builds #22 and #25 here ran every suite because their own messages quoted it. GHA behaves the same way.

## Test plan
- [x] `cli-argument-parsing` / `cli-error-handling` pass on an NPU pod (build #22, before they were gated)
- [ ] An ordinary PR uploads 8 path-gated steps at the default level and no advanced CLI steps
- [ ] A `[pytest-full]` commit uploads 10 steps at the full level with no path gating


rebel-junamsong (approved):
LGTM

PR #728, opened 2026-09-09
other: ci - run PR pytest suites on Obedients alongside GHA

rebel-thkim:
## Summary
Phase 1 of moving optimum-rbln CI off GitHub Actions onto Obedients, following the vllm-rbln layout (#882 / #931 there).

- `.buildkite/pr.yml`: team-member gate, then uploads the pytest pipeline with `--git-diff-base` (ready for `if_changed` in a later phase).
- `.buildkite/pipelines/pytest/pipeline.yml`: one step per GHA matrix entry — `config`, `cli-basic`, `transformers`, `diffusers`, `llm` 1–4, each on 1× RBLN-CA25 (`config`/`cli-basic` compile without an explicit `npu`, so they need an NPU pod too). Reports `[OB] Optimum-RBLN pytest` as a commit status.
- `scripts/install-rebel-compiler.sh` installs the version pinned in `.github/version.yaml` from the cluster's internal PyPI (`REBEL_PYPI_INTERNAL_ENDPOINT`), so both CIs test the same compiler.
- The sync step writes `src/optimum/rbln/__version__.py` from the installed metadata when the hatch-vcs hook left none: pods share `UV_CACHE_DIR`, so concurrent steps building the same commit reuse one pod's editable build and only that pod gets the generated file (seen in builds #2–#4).
- `scripts/run-suite.sh` runs one suite and honors the same `[skip-transformers]` / `[skip-diffusers]` / `[skip-llms]` / `[skip-cli]` directives via `BUILDKITE_MESSAGE`.

GHA pytest is untouched in this PR; it stays on until Obedients results match on the same commits. Deploy, compiler-update bot, renovate and PR-title check stay on GHA.

Working doc (phases, mapping, open items): internal artifact, ask thkim.

## Test plan
- [x] Obedients build passes gate → sync → rebel-compiler install → all 8 suites ([build #5](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-pr-ci/builds/5), 18 min incl. queue)
- [x] `config` / `cli-basic` on CPU pods: not possible (compile needs an NPU) → moved to NPU
- [ ] Suite results match GHA on the same commit (3+ PRs) — 1/3: all 10 GHA suites and all 8 OB suites pass on this branch
- [ ] `[skip-llms]` commit skips the llm steps
- [ ] Webhook builds: PR open / push did not create OB builds yet; builds #1–#5 were triggered manually

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #742, opened 2026-09-13
other: ci - let a compiler build drive this pipeline

rebel-junamsong:
Prerequisite for adding optimum-rbln to rebel_compiler's `:rocket: Downstream CI`
group (rebellions-sw/rebel_compiler#13567). Without it that gate would be green
and meaningless — it would run zero suites against the pinned compiler.

## What was in the way

**The compiler version was not overridable.** `install-rebel-compiler.sh` reads
the pin from `.github/version.yaml` unconditionally, so a build triggered at
`rebel-compiler==<the wheel we just built>` would install the pinned one instead
and report on it. `REBEL_COMPILER_VERSION` now takes precedence when set, and the
step annotates what it installed; with the variable unset nothing changes. This
is the same contract `vllm-rbln`'s `rebel-compiler-override.sh` already offers to
the same trigger.

**The suites path-gated themselves away.** Both uploads pass
`--git-diff-base origin/$BUILDKITE_PULL_REQUEST_BASE_BRANCH`, which falls back to
`dev`. On a build with no pull request that base *is* the build's own commit, so
the diff is empty and every `if_changed` suite is skipped — the build passes
having run nothing. A non-PR build (the compiler trigger, or a manual run) now
uploads without a diff base, exactly as the `[pytest-full]` path already does.
PR builds are untouched.

## Verified

Rendered both upload commands out of `pr.yml` (`$$` → `$`) and ran them against
stub `buildkite-agent`/`git`:

| build | pytest upload | BC upload |
|---|---|---|
| PR #728 | `--git-diff-base origin/dev` | `--git-diff-base origin/dev` |
| no PR (trigger/manual) | no diff base | no diff base |
| `[pytest-full]` | no diff base (unchanged) | no diff base (unchanged) |

And the installer against a stub `uv`: unset → `rebel-compiler==0.11.3.dev58+gf0e2aad4.prod`
(the current pin), set → the version passed, plus the annotation.


rebel-thkim (approved):
👍

PR #743, opened 2026-09-14
other(test): pin push_to_hub read-back to the pushed commit

rebel-thkim:
## Problem

`TestSDModel::test_push_to_hub` failed in [nightly build 6](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-nightly-ci/builds/6?job=5981680b-d8cb-4b4b-9632-32e53b327a48):

```
FAILED tests/test_diffusers.py::TestSDModel::test_push_to_hub
  AssertionError: 196099564871151726944940150660443205695
              != 226174774841618392348414164106024107152
```

The test writes a random 128-bit hash into `text_encoder.config.from_local`, pushes it, and reads it back by resolving `main`.

This is a read-after-write race, not a product regression. Confirmed against the Hub:

- The push created commit `8164550c` on `Rebellions/rbln-tiny-sd-pipe` at **22:01:14 UTC**
- The job started at 22:00:32 and recorded the failure at ~22:01:16, i.e. the read happened 1-2s after the commit landed
- The value it read back belongs to the repo's previous commit (same day, 08:54:45)

So the upload succeeded and the read was served the previous revision. The same test failed identically in build 3 and passed in build 4.

## Fix

Resolve the sha of the commit the push just created and read back at that revision. Commit-pinned paths are immutable, so the round-trip no longer depends on `main` having propagated, and a concurrent push from another job can no longer clobber the value under us.

The assertion is not weakened:

- push creates no commit -> sha does not move -> old value -> still fails
- commit exists but is missing the modified config -> still fails

Also drops `private=True` from the two `from_pretrained` calls. It is not a parameter there and only ended up as an unused config attribute; the `save_pretrained` one is real and is kept.

## Test plan

CI. The change affects both the diffusers (`CLIPConfig`) and transformers (`AutoConfig`) branches of `BaseHubTest.test_push_to_hub`.


tests/test_base.py
```diff
@@ -160,32 +161,34 @@ def test_push_to_hub(self):
...
+                # a commit can still serve the previous revision.
+                revision = HfApi(token=HF_AUTH_TOKEN).repo_info(repo_id).sha
+
+                # AutoConfig rather than CLIPConfig for the text encoder: the inherited
+                # PreTrainedConfig.from_pretrained takes `token` as a named argument and
+                # never puts it back into the kwargs it forwards to get_config_dict
+                # (transformers 5.15.1), so the read goes out anonymous, the private repo
+                # answers 404, and huggingface_hub serves whatever the shared cache holds.
+                # AutoConfig forwards the token, and resolves the checkpoint to the
+                # CLIPTextConfig it actually is.
```
rebel-jongho: please trim comment

rebel-jongho (approved):

PR #744, opened 2026-09-14
other: ci - balance the pytest shards by class runtime

rebel-thkim:
## Problem

From [nightly build 6](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-nightly-ci/builds/6):

| job | duration |
|---|---|
| `llm 4/4` | 8m56 |
| `transformers` | 8m55 |
| `llm 3/4` | 8m20 |
| `diffusers` | 5m35 |
| `llm 1/4`, `llm 2/4` | 3m13, 3m07 |

Two causes behind the llm spread:

1. **`pytest-split` had no `.test_durations` to read**, so it fell back to splitting by test count. Test count is a poor proxy here — each class compiles its model in `setUpClass`, and those setups ranged from 6s to 72s.
2. **Splitting by test cuts through a class.** `TestQwen3VLForConditionalGeneration::test_automap` ran in `llm 3/4` and `::test_propagate_config` in `llm 4/4`, so that class's compile ran twice — on the two slowest shards.

`transformers` and `diffusers` were not split at all.

## Change

Shard by class, inside pytest. `tests/conftest.py` adds `--shard INDEX/COUNT` and filters the collected items in `pytest_collection_modifyitems`:

```python
classes = list(dict.fromkeys(item.cls for item in items))
keep = set(classes[index - 1 :: count])
items[:] = [item for item in items if item.cls in keep]
```

- **By class**, so a model never compiles in two jobs at once.
- **Round-robin**, so there is no cost table to keep up to date. Adding or removing a test changes nothing here. It balances because the expensive model families are defined next to each other in the file, so taking every nth class spreads them.
- **Order preserved.** `[i-1::n]` leaves each shard in collection order, which matters: the suite is not order-independent, and an earlier revision of this PR turned `TestGroundingDinoModel` into the first class of a shard and hit the `HF_CONFIG_KWARGS` leak that #741 later fixed. Splitting can only remove an ordering interaction this way, never introduce one.
- **Inside pytest**, so there is no `--collect-only` pre-pass and no node ids passed through the shell. `run-suite.sh` goes back to one line per suite.

Counts live only in the step arguments, character-for-character identical to the label, so a mismatch is visible on sight:

```yaml
- label: ":pytest: llm 3/6"
  ... "bash .buildkite/scripts/run-suite.sh llm 3/6"
```

llm 4 → 6, transformers 1 → 3, diffusers 1 → 2.

## On the numbers

The counts are chosen with headroom rather than fitted to a measurement, because wall clock is not comparable between builds on this cluster.

[Build 77](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-pr-ci/builds/77) ran green at `[pytest-full]` but 1.44x slower than the nightly baseline on `diffusers` — a job this PR does not touch, on the same code and the same compiler (`rebel-compiler==0.11.3.dev58+gf0e2aad4.prod`): 5m35 → 8m01. Sizing shards against a single nightly's `--durations`, as an earlier revision of this PR did, produced per-class seconds that were a snapshot of one node on one evening.

Projected from the nightly baseline with #745 applied, worst shard: llm 4m53, transformers 3m59, diffusers 3m10. That leaves roughly the 1.4x a busy day costs before any of them reaches seven minutes.

Four more pods than dev runs today. The structural part does not depend on the numbers: no duplicate `setUpClass` compiles, and `transformers` and `diffusers` get split at all.

## Depends on

#745 took `test_propagate_config` off the real checkpoint, which dropped the heaviest class in `test_llm.py` from 164s to 53s. Without it, no shard count fixes llm — one class was 2m43 on its own.

## Test plan

[Build 77](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-pr-ci/builds/77) passed every suite at `[pytest-full]` with the sharding applied, which is the correctness check. The shard counts have since been raised; the durations that motivated them are in the section above.


rebel-junamsong (approved):
LGTM

PR #745, opened 2026-09-14
other(llm): stop test_propagate_config compiling the real checkpoint

rebel-thkim:
## Problem

`setUpClass` in these VLM test classes shrinks the model before compiling — it edits `vision_config` and `text_config` down to one or two layers and stores them in `HF_CONFIG_KWARGS`, which the base passes to `from_pretrained` (`tests/test_base.py:234`).

`test_propagate_config` calls `from_pretrained` itself and passed only `RBLN_CLASS_KWARGS`:

```python
self.RBLN_CLASS_KWARGS["rbln_config"].update({"create_runtimes": False})
model = self.RBLN_CLASS.from_pretrained(model_id=self.HF_MODEL_ID, **self.RBLN_CLASS_KWARGS)
#                                                                  ^ no HF_CONFIG_KWARGS
```

So it was the one test in the class compiling the model as published. `create_runtimes=False` skips runtime creation, not the compile.

Measured on [nightly build 6](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-nightly-ci/builds/6), `--durations 0`:

| class | `HF_MODEL_ID` | `test_propagate_config` |
|---|---|---|
| `TestQwen2_5_VLForConditionalGeneration` | `Qwen/Qwen2.5-VL-3B-Instruct` | **133.45s** |
| `TestQwen3VLForConditionalGeneration` | `Qwen/Qwen3-VL-2B-Instruct` | **86.96s** |
| `TestQwen2VLForConditionalGeneration` | `hf-internal-testing/tiny-random-…` | 16.37s |
| `TestLlavaNextForConditionalGeneration` | `trl-internal-testing/tiny-…` | 15.10s |

133s of that is a single test asserting two booleans, and it is 82% of `TestQwen2_5_VLForConditionalGeneration`'s 164s.

The two tiny-checkpoint classes are cheap, but `TestLlavaNextForConditionalGeneration` pins a revision in `HF_CONFIG_KWARGS`; dropping it meant this one test resolved the model at HEAD while every other test in the class used the pin.

## Fix

Pass `HF_CONFIG_KWARGS`, so the test compiles the same stub the rest of the class does.

Also build the `rbln_config` as a new dict instead of updating the class attribute in place. `create_runtimes=False` was staying in `RBLN_CLASS_KWARGS` for the remainder of the class, and `test_save_artifacts` and `test_save_load` sort after `test_propagate_config` — the same shared-mutable-class-attribute shape as #741.

## Why it matters beyond the four tests

This is the reason the pytest shards are hard to balance. `TestQwen2_5_VLForConditionalGeneration` is the heaviest class in `test_llm.py` at 164s, and 133s of that is this test. Without it the class is ~30s, and the heaviest class in the suite drops to `TestBlip2ForConditionalGeneration` at 53s — small enough that the split stops needing to be clever. #744 is on hold behind this.

## Test plan

CI at `[pytest-full]`, comparing the `--durations 0` entries for these four tests against the numbers above.


rebel-seinpark (approved):
LGTM!

PR #559, opened 2026-05-11
model: decoderonly batch_attn_decode opt

rebel-thkim:
# Pull Request Description

> **⚠️ Important: Branch Target**
> - **New features, enhancements, and non-critical fixes**: Merge to `dev` branch
> - **Critical hotfixes only**: Merge to `main` branch (must also merge to `dev`)
> 
> Please ensure you've selected the correct base branch before submitting!

## Type of Change

- [ ] Release (dev → main merge for production release)
- [ ] New Model Support
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please describe):

## Related Issues

----
# Conventional commit
```
type(optional scope): description
```
----
# Type candidate
  - Model Updates
    - `model`: Adding New models or Bugfix for existing models
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
    - `performance`: Optimizing some models or this library itself
      - ex) Loading RBLNModel faster
      - ex) Optimizing Memory Usage of DecoderOnlyModel
  - Code Refactor
    - `refactor`: Re-arrange class architecture, or more.
      - ex) Refactor Seq2Seq
  - Documentation
    - `doc`: Update docstring only
  - Library Dependencies
    - `dependency`: Update requirements, something like that.
  - Release
    - `release`: Merging dev to main for production release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update

PR #746, opened 2026-09-14
other: tests - run the batch-sort tests on a cpu-only step

rebel-thkim:
## Problem

Three test files have never been run by anything:

| file | tests |
|---|---|
| `tests/test_batch_attn_sort.py` | 15 |
| `tests/test_batch_sort_vlm_composition.py` | 14 |
| `tests/test_batch_sort_vlm_flatten.py` | 13 |

They arrived together in `825fe83d` (#559). Nothing references them — `.buildkite/**` and `.github/workflows/**` name only `test_config.py`, `test_transformers.py`, `test_diffusers.py`, `test_llm.py` and `test_cli.py`, so 42 tests have been sitting in the repo as dead weight.

They cover the host side of the batch-attn sort: tie-break stability, the unsort round-trip, per-model vision-input permutation for ten VLM families, and six paths that must raise. None of that is reachable from the end-to-end model tests.

## They are not NPU tests

No `from_pretrained`, no compile, no runtime creation. They drive the sort with `monkeypatch` and fakes (`_FakeCausalLM`, `SimpleNamespace`), and call `npu_is_cr13_or_later("RBLN-CR13")` with a string rather than querying a device.

So porting them into `test_llm.py` — the other option on the table — would be wrong twice over: every class there compiles a model and holds an NPU, and the suite is split across shards, so these would scatter and pay for a device they never touch.

## Change

Move them to `tests/unit/cpu/` and give that directory a step:

```yaml
- label: ":pytest: unit-cpu"
  if_changed: *if_changed_unit_cpu
  image: "${DEVTOOLS_DOCKER_IMAGE}"
  secrets: *secrets
  timeout_in_minutes: 60
  command:
    - *sync
    - "bash .buildkite/scripts/run-suite.sh unit-cpu"
```

```bash
unit-cpu)
  uv run --no-sync pytest tests/unit/cpu -vv --durations 0 ;;
```

**The directory is the list.** A new test is picked up by dropping a file in `tests/unit/cpu/`. There is nothing to enumerate in the pipeline and keep in sync — the failure mode that put these three files in this state to begin with.

**No `resources: *npu`.** These never touch a device, so the step should not reserve one. Without the request the device plugin does not inject `/dev/rbln*`, so a test that does need a device fails loudly rather than quietly holding hardware.

**Two levels, because there are two axes.** `unit` is the scope; `cpu` is what the job needs, which is what the step branches on. Folding them into one name drops an axis: a unit test that *does* need a device has no home under `tests/cpu/`, and under `tests/unit/` it would belong there by name but not by step — which ends with the step taking the resource and the property going away for every other test in it. Nested, that test goes to `tests/unit/npu/` and nothing here has to change.

## Do they pass?

Not known. They have never run, and the code has moved since #559 — the `transformers` 5.15.1 bump among other things. Verifying locally is not possible in this checkout: the project venv is Python 3.13 while `rebel-compiler` ships 3.12 binaries, so `import rebel` fails on the ABI.

That is the point of running them first. Whether all 42 earn their place is a fair question, and the answer varies — `test_qwen3_vl_moe_inherits_mixin` asserts a mixin and one flag, while `test_requires_batch_sort_not_user_settable` pins a derived config field against both the constructor and the attribute path. But pruning coverage that has never executed tells you nothing about what you removed, so: run them, then decide per test.

## Test plan

The first build answers two things — whether the 42 pass, and whether `import optimum.rbln` works on a step with no NPU reserved. If the import needs a device, the step gets `resources: *npu` back; if some tests fail, the follow-up is per test, fix or drop with a reason.

A tight `timeout_in_minutes` on this step is worth adding once there is a measured duration to base it on. Nothing enforces that a suite named for its lack of a device stays fast — compilation does not need one either — and a wall-clock bound is the cheap guard against a slow test drifting in.


rebel-junamsong (approved):
LGTM

PR #750, opened 2026-09-15
model(qwen2_5_vl): skip loading the visual encoder with _load_visual_runtime

rebel-seinpark:
## Description

Ports the Qwen3-VL `_load_visual_runtime` mechanism to Qwen2.5-VL, so a text-only consumer can skip **loading** the visual encoder entirely — no compiled-model (.rbln) read, no torch artifacts, no host memory for a submodule that is never called.

Motivation: the Cosmos-Predict2.5 / Transfer2.5 pipelines use `RBLNQwen2_5_VLForConditionalGeneration` as their text encoder with text-only prompts. The vision tower is a mandatory submodule, so today it is compiled (fine — minimal window) and then **read back into host memory on every load** even though `visual: {create_runtimes: False}` skips its runtimes; on the 7B checkpoint that is a ~1.3 GB blob and the corresponding load time, for a module that can never run.

### Changes

- `RBLNQwen2_5_VLForConditionalGenerationConfig` / `RBLNQwen2_5_VLModelConfig`: add `_load_visual_runtime: bool = True` (a `subclass_non_save_attributes` entry, mirroring Qwen3-VL — runtime-only, never serialized).
- `RBLNQwen2_5_VLModel._load_submodules`: when loading compiled artifacts with the flag off, return no submodules (same three-liner as Qwen3-VL). Export is unaffected — the visual encoder is still compiled and saved.
- `self.visual = self.rbln_submodules[0]` → safe lookup (`None` when skipped); `pixel_values` / `pixel_values_videos` paths raise a `RuntimeError` with guidance when the visual encoder is not loaded.

### Verification

On the tiny Cosmos text encoder (`rbln/tiny-cosmos-2.5-predict`, NPU):
- flag on (default): visual submodule loads exactly as before
- flag off: `visual is None`, load succeeds without touching the visual artifacts, and the text `hidden_states` are **bit-identical** to the flag-on run
- pixel inputs with the flag off raise the guidance error

The default path is behavior-preserving; the Qwen2.5-VL CI suite (3B) covers the VLM regression.

### Follow-up

The Cosmos pipeline configs will default their text encoder to `_load_visual_runtime=False` on the Cosmos branch (#694). Skipping the visual **compile** for text-only exports is a separate, more invasive follow-up (mandatory-submodule structure).


rebel-seinpark:
@rebel-jongho 
- `create_runtimes=False` = "not running it *now*, but may run it later". It only saves device memory; the host costs (disk I/O + checksum verification + RAM for the compiled blob) are still paid.
- `_load_visual_runtime=False` = "never running it" — the host costs drop to zero as well. The right fit for consumers that never call the vision tower, such as the Cosmos text encoder.

src/optimum/rbln/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
```diff
@@ -368,6 +368,14 @@ class RBLNQwen2_5_VLModel(RBLNDecoderOnlyModel):
     _get_rope_index_func = Qwen2_5_VLModel.get_rope_index
     get_vision_position_ids = Qwen2_5_VLModel.get_vision_position_ids
 
+    @classmethod
+    def _load_submodules(cls, model_save_dir, rbln_config, model=None, **kwargs):
+        # Loading with _load_visual_runtime=False skips the visual encoder entirely (no
+        # compiled-model read, no torch artifacts); text-only callers never miss it.
```
rebel-seinpark:
```suggestion
```

rebel-jongho (commented):
1. subclass_non_save_attributes silently drops memory_budget
2. Design: let create_runtimes=False skip reading the compiled models

      Stepping back — I think we already have the knob for this, it just does less than its users expect. On the load path, create_runtimes=False still deserializes every .rbln into host memory and only then declines to allocate on the NPU. Nobody asks for that: if you are not going to run the model, reading the blob is pure cost. Making create_runtimes=False skip the read outright gives this PR's behaviour with no new API at all.

      Concretely, visual={"max_seq_len": 64, "create_runtimes": False} — which is exactly what the Cosmos branch wrote before this PR — would simply start doing what it says, and _load_visual_runtime would not be needed.

      The split is already clean in the code, so this is essentially one guard:

      - _load_compiled_models() is called from a single place, modeling_base.py:309, and that whole block sits inside if rbln_compiled_models is None: at modeling_base.py:234. The export path passes the compiled models in explicitly, so export=True, create_runtimes=False — the compile-only workflow the CLI defaults to at cli.py:790 — is untouched by construction.
      - UnavailableRuntime (utils/runtime_utils.py:208-256) already carries the "loaded without runtimes" story: __getitem__ returns self, and __call__ raises a message telling you to reload with create_runtimes=True. That is why RBLNQwen2_5_VisionTransformerPretrainedModel.__post_init__'s self.transformer = self.model[0] keeps working today, and why the Cosmos config worked before this PR.
      - Per-submodule granularity already exists: create_runtimes propagates parent → child, but an explicit value on the child wins.

      What this buys over the flag in this PR:

      - No new config attribute, so no subclass_non_save_attributes entry, so the problem in point 1 never arises.
      - The submodule list keeps its length. This is the part I care about most. rbln_submodules is indexed positionally everywhere — self.language_model = self.rbln_submodules[1] in gemma3 / gemma4 / llava / llava_next / paligemma, [2] in blip_2, [3] in grounding_dino — and __repr__ (modeling_base.py:629) pairs entry i with _rbln_submodules[i]["name"]. Any design that shortens the list only works while visual is the sole submodule; this one sidesteps the question entirely and extends to qwen3_asr's audio_tower, gemma3's vision_tower and so on for free.
      - vllm-rbln can drop _load_visual_runtime and pass {"visual": {"create_runtimes": False}} instead (vllm_rbln/model_executor/models/optimum/model_base.py:213). While we are there, the hardcoded ec_enabled_model = model_cls_name == "RBLNQwen3VLForConditionalGeneration" at line 195 of the same file could go too.

      I checked who actually reads compiled_models after a load, and the blast radius is close to empty:

      - cli.py:466 (kvcache resize) reads the .rbln files directly with rebel.RBLNCompiledModel(p) and never goes through from_pretrained.
      - diffusers/modeling_diffusers.py:477-480 aggregates submodule.compiled_models, but _construct_pipe is export-only — from_pretrained returns early at modeling_diffusers.py:277 (if not export: return model).
      - tests/test_config.py:53 asserts on compiled_models[0]._meta, but from an export=True fixture.
      - multicontrolnet.py:41 exposes compiled_models as a property; it would only be affected if something read it after a runtime-less load.

      Three things I would want handled:

      1. Keep the artifact-integrity check. _load_compiled_models raises FileNotFoundError for a missing .rbln at modeling_base.py:206-210, and tests/test_base.py:324-349 (_inner_test_save_load) runs the whole save → load round trip under ContextRblnConfig(create_runtimes=False). That is currently our main check that save_pretrained actually copied the binaries. Keep the existence/name validation and skip only the rebel.RBLNCompiledModel(...) construction.
      2. Guard modeling_base.py:349-350 — compiled_models = [rbln_compiled_models[cm_name] for cm_name in compiled_model_names] would KeyError on an empty mapping.
      3. The "is it there?" test changes for qwen3_vl. It currently checks self.visual is None; with this approach self.visual is a live object holding an UnavailableRuntime, so the disaggregation path would surface "reload with create_runtimes=True" instead of "provide pre-computed image_embeds". A small has_runtime property on RBLNBaseModel (not isinstance(self.model, UnavailableRuntime)) would give both call sites something honest to branch on.

      The one thing the two approaches do not share: with create_runtimes=False the submodule object is still constructed, so host-side torch artifacts still load. For the Qwen2.5-VL vision encoder that is just patch_embed out of torch_artifacts.pth, a few MB, so it does not matter here — I would not design for it until some model makes it hurt.

      create_runtimes would end up meaning something slightly wider than its name (no runtimes and no artifact binaries read). I think that is fine, but the docstring at configuration_utils.py:501-507 should be updated to say so.

rebel-thkim:
I also think the approach Jong-ho suggested looks better. I'm actually curious about how to use it—is there a way to see some sample code?

rebel-seinpark:
It already used in disagg-encoder. It merged from https://github.com/RBLN-SW/optimum-rbln/pull/551

rebel-seinpark:
https://github.com/RBLN-SW/optimum-rbln/pull/760#pullrequestreview-5231304754

PR #736, opened 2026-09-10
other: add mypy type checking for src and run it in CI

rebel-jongho:
# Pull Request Description

> **⚠️ Important: Branch Target**
> - **New features, enhancements, and non-critical fixes**: Merge to `dev` branch
> - **Critical hotfixes only**: Merge to `main` branch (must also merge to `dev`)

## Type of Change
- [ ] Release (dev → main merge for production release)
- [ ] New Model Support
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please describe):

## Changes Overview

This PR introduces static type checking with **mypy** and makes it pass on `src/`.

- **Configuration**: `[tool.mypy]` in `pyproject.toml` (Python 3.10 target, src layout, `check_untyped_defs`, `no_implicit_optional`, `warn_unused_ignores`, `warn_redundant_casts`). Untyped third parties (`rebel`, `torchvision`, `retinaface`, `cosmos_guardrail`) are scoped with `ignore_missing_imports` per module instead of globally, so torch/transformers/diffusers types stay real.
- **`py.typed`** is added to `optimum/rbln` and ships in the wheel, so downstream packages (e.g. vllm-rbln) can consume our annotations.
- **CI**: `uv run --no-sync mypy` runs next to ruff in `check_code_quality.yml` and in the Buildkite `code quality` gate (`.buildkite/pr.yml`). `mypy` is added to the `quality` dependency group; `typing-extensions` (already an indirect dependency) becomes explicit for `Self`.
- **Core modules are fully typed** (`modeling_base.py`, `modeling.py`, `configuration_utils.py`, `modeling_rope_utils.py`, `cli.py`, `utils/*`, `transformers/modeling_generic.py`, `transformers/utils/*`, `transformers/cache_utils.py`, `transformers/modeling_attention_utils.py`, `diffusers/modeling_diffusers.py`, `diffusers/configurations/*`). Highlights:
  - `RBLNBaseModel` declares the class-level contract subclasses fill in (`_hf_class`, `_rbln_config_class`, `auto_model_class`, and `get_pytorch_model` / `from_model` / `_create_runtimes` / `_update_rbln_config` / `forward` stubs), and `from_pretrained` / `from_model` / `_from_pretrained` return `Self`.
[...]

PR #751, opened 2026-09-15
performance: keep quantized weights as safetensors mmap views instead of copying them

rebel-jongho:
# Pull Request Description

>

## Type of Change
- [ ] New Model Support
- [ ] Documentation update
- [x] Performance improvement
- [ ] Code refactoring

## Changes Overview

`load_weights_from_files` in `rbln_quantization.py` no longer copies checkpoint tensors into the model's parameters when the dtype and shape already match. The parameter adopts the safetensors mmap view instead (`target.data = value`); tensors that `canonicalize_checkpoint_items` reshapes or that need a dtype cast (the weight, input and kv scales) keep the cast-and-copy path. The parameter/buffer lookup happens once, so the two former copy blocks collapse into a single branch.

For an fp8 checkpoint every quantized weight takes the adopt branch, since `convert_to_qfloat_linear` creates the parameter with the checkpoint's dtype (`float8_e4m3fn`) and shape.

## Motivation and Context

While compiling Llama3.3-70B-FP8 (`amd/Llama-3.3-70B-Instruct-FP8-KV`, 67.7 GiB) the host peak is 192.4 GiB, above the 175 GiB budget. The kernel pool and init blobs are compiler artifacts that must stay resident; the only removable piece is a 68.1 GiB anonymous copy of the checkpoint that the quantized loader makes with `copy_()` while the safetensors mapping of the same bytes is still alive. Dense models do not have this problem because transformers leaves their weights as mmap views, and MoE models release the mapping deliberately in `RBLNMoeLoadMixin`; the quantized decoder-only path was the one place that copied.

Measured on the 70B case (compiler unchanged, cgroup `memory.stat` sampled): anon peak 192.4 → 126.0 GiB, compile wall time −20%, `WEIGHT_POOL_STAT` identical before and after. Details are in the internal analysis shared in the host-memory thread.

### Local checks

A synthetic fp8 checkpoint (4-layer Llama, hidden 2048, fp8 weights with fp32 weight/k/v scales, 138 MiB) loaded through `get_quantized_model`:

| | before | after |
| --- | --- | --- |
| RssAnon growth during load | +143 MiB | +7 MiB |
| weights equal to checkpoint bytes, scales loaded | yes | yes |
| CPU forward logits | reference | bit-identical |

Compiling the same checkpoint with `RBLNLlamaForCausalLM.from_pretrained(export=True, rbln_create_runtimes=False)` and the settings of `TestLlamaForCausalLM_fp8` (RBLN-CR03, flash_attn, partition 4096) succeeds before and after; the anon growth over the whole compile drops from 848 to 661 MiB. The `.rbln` files were not compared byte for byte: two compiles of the unchanged code already differ in ~35 KB of metadata (and by 16 bytes in size), and the before/after difference has the same size and location, so the weight payload is unchanged as far as a byte comparison can tell.

### Notes for review

- The mapping is kept alive by the tensors themselves (verified: after adopting and dropping the `load_file` dicts the parameters stay readable and RssAnon does not grow). The mapping is writable, so any in-place write to a weight would silently copy-on-write that page back to anonymous memory; nothing on the compile path writes to parameters after loading (`normalize_contiguous_` only replaces non-contiguous tensors, and safetensors views are contiguous).
- MoE checkpoints are unaffected: `RBLNMoeLoadMixin.release_checkpoint_mmap_` clones every parameter right after loading.
- Because parameters are now file-backed, `VmRSS` (anon + file) as read by the compiler's `currentProcessRSS()` changes shape: the first `WEIGHT_POOL_PLAN` reports a lower `rss_gib` and picks `mode=fast` instead of `mode=save`. Whether that plan switch is desired is a compiler-side question.
- #736 touches the same two lines; whichever lands second needs a trivial conflict resolution.

## Related Issues

None.


rebel-thkim (approved):
👍

PR #752, opened 2026-09-15
other: ci - restore the [slack-report] step on PR builds

rebel-thkim:
## What

`[slack-report]` in a commit message no longer does anything on a PR build.

Before #737, `rbln_trigger_on_pr.yaml` had a job for it:

```yaml
  slack-report:
    if: needs.check-slack-report.outputs.should_slack_report == 'true'
    uses: ./.github/workflows/slack_report.yaml
    with:
      title_base: "Optimum-RBLN PR Full Test Results"
      use_playground_channel: true
```

#737 deleted that workflow and carried the report over to `nightly.yml` only — `pr.yml` never got one. Meanwhile `renovate.json` and `auto_compiler_update.yaml` still prefix every compiler-update PR with `[pytest-full] [slack-report]` (e.g. [PR CI #101](https://obedients.k8s.rebellions.in/orgs/main/pipelines/optimum-rbln-pr-ci/builds/101)), so those PRs have been silently reporting nothing.

## How

The step goes at the end of `pr.yml`, behind a second `wait`, matching what the nightly does:

```yaml
  - wait: ~
    continue_on_failure: true

  - label: ":slack: report"
    if: build.message =~ /\[slack-report\]/
    soft_fail: true
    allow_dependency_failure: true
```

`continue_on_failure` is a sibling of `wait`, not a value under it — nested, the flag is dropped and a failed suite takes the report down with it (nightly build #1).

| | GHA | here |
|---|---|---|
| trigger | `[slack-report]` in the head commit message | `if: build.message =~ /\[slack-report\]/` |
| channel | `use_playground_channel: true` | `slack-report.py` already picks the playground channel for any build whose source is not `schedule` |
| title | `Optimum-RBLN PR Full Test Results` | `Optimum-RBLN PR Test Results` |

The title drops "Full". `[slack-report]` and `[pytest-full]` are independent markers — the GHA job knew this (it passed `'default'` for the versions when `should_full_test` was false) but still titled every report "Full Test Results", so a default-level run reported itself as a full one.

Without the marker the step is not created, so no agent is held. It carries no `github_commit_status`, so it cannot affect a required check.

## Verification

`pr.yml` parses; the report is the last of 8 steps, after the new `wait`.

Delivery itself can only be confirmed on a build whose commit message carries the marker — the next compiler-update PR, or a commit made for the purpose.


rebel-junamsong (approved):
LGTM

PR #760, opened 2026-09-16
performance: read compiled models lazily so create_runtimes=False skips the .rbln blobs; deprecate _load_visual_runtime

rebel-jongho:
## Type of Change
- [x] Performance improvement
- [x] Code refactoring

## Changes Overview

**`create_runtimes=False` no longer reads the compiled models from disk.** `RBLNBaseModel.compiled_models` is now a lazy property: the load path only resolves and validates the `.rbln` paths, and the blobs are read on first access (`save_pretrained`, the CLI and `_construct_pipe` keep working; a model that never touches `compiled_models` never pays for them). The compile path is unchanged and still hands the freshly compiled objects to the model.

Because runtime options already propagate per submodule, this gives every composite model a uniform way to skip a submodule: `rbln_config={"visual": {"create_runtimes": False}}` loads the language model normally and leaves the vision encoder as an `UnavailableRuntime` with nothing read from disk.

**`_load_visual_runtime` is deprecated** on the Qwen3-VL and Qwen3.5 configs in favour of that spelling.
- The `_load_submodules` overrides that returned an empty submodule list are removed, so `rbln_submodules` keeps its positional layout and `self.visual` is always present. Calling it without a runtime raises the standard `UnavailableRuntime` error.
- The knob is kept working through 0.11.x with a deprecation warning, via the existing decorators:
  - `@deprecate_kwarg` on the four `__init__`s. The parameter is gone from the signatures; as a constructor kwarg it never had an effect (the old override only fired on the load path, and a config *instance* passed to `from_pretrained` did not carry it either), so warn-and-drop is behaviour-preserving.
  - `DeprecatedLoadVisualRuntimeMixin` gives the configs a `_load_visual_runtime` property whose setter (`@deprecate_method`) forwards to `visual.create_runtimes`. This is the path that actually matters: `from_pretrained(path, rbln_config={"_load_visual_runtime": False})` applies non-save keys with `setattr` after construction (`configuration_utils.py`, `from_pretrained`), so a decorator on `__init__` alone would never see it. The mixin lives in the Qwen3-VL config module and is imported by Qwen3.5.
  - Both decorators use `raise_if_greater_or_equal_version=False`, like `tensor_parallel_size`, because the one real consumer is vllm-rbln's disaggregated-encoder consumer node.

## Motivation and Context

`rebel.RBLNCompiledModel(path)` copies the whole file into anonymous host memory (no mmap), so `from_pretrained(..., create_runtimes=False)` was paying the full blob cost for nothing. Measured on the OPT test artifact (907 MiB of `.rbln`):

| | before | after |
|---|---|---|
| `from_pretrained(create_runtimes=False)` | +918 MiB, 0.69 s | +4 MiB, 0.01 s |
| first `model.compiled_models` access | – | +916 MiB, 0.70 s |

For the disaggregated-encoder use case (#551) the same mechanism now covers Qwen2.5-VL (whose visual blob is 1.3 GiB for the 3B model) and any future composite model without a per-model flag; #750 proposed adding `_load_visual_runtime` to Qwen2.5-VL and is superseded by this.

Verified end to end on a 1-text-layer Qwen3-VL-2B artifact (`visual/compiled_model.rbln` = 794 MiB): `visual={"create_runtimes": False}` and the deprecated `_load_visual_runtime=False` both create the LLM runtimes on the NPU while the visual encoder's blob is never read (`visual._compiled_models is None`, `visual.model` is an `UnavailableRuntime`, and calling it or passing `pixel_values` raises the standard `create_runtimes=False` error). `load_visual_encoder` and a full load with image generation are unchanged.

## Related Issues

Supersedes #750. Follows up #551.


src/optimum/rbln/modeling_base.py
```diff
@@ -306,7 +330,10 @@ class DummyConfigMixin(ConfigMixin):
                     config = PretrainedConfig(**config_dict)
 
             compiled_model_names = [cfg.compiled_model_name for cfg in rbln_config.compile_cfgs]
-            rbln_compiled_models = cls._load_compiled_models(model_path_subfolder, compiled_model_names)
+            if rbln_config.create_runtimes:
+                rbln_compiled_models = cls._load_compiled_models(model_path_subfolder, compiled_model_names)
+            else:
+                cls._resolve_compiled_model_paths(model_path_subfolder, compiled_model_names)
```
rebel-thkim: Since `return` isn't used anywhere and `_resolve_compiled_model_paths` itself doesn't seem to update any attributes, isn't the `else` statement dead code?

rebel-jongho: Not dead, but it certainly read like it. The call was there for its side effect: `_resolve_compiled_model_paths` raises `FileNotFoundError` when an expected `.rbln` is missing (and warns on unexpected ones), so a `create_runtimes=False` load still fails fast on a broken artifact, which is what `tests/test_base.py`'s save/load round-trip relies on.

Restructured in 04e99dee so the intent is in the code: the paths are resolved unconditionally and only the read is gated on `create_runtimes`. No discarded return value anymore.

src/optimum/rbln/transformers/models/qwen3_vl/configuration_qwen3_vl.py
```diff
@@ -15,10 +15,35 @@
 from typing import Any
 
 from ....configuration_utils import RBLNModelConfig
+from ....utils.deprecation import deprecate_kwarg, deprecate_method
 from ..decoderonly.configuration_decoderonly import RBLNDecoderOnlyModelConfig, RBLNDecoderOnlyModelForCausalLMConfig
 
 
-class RBLNQwen3VLForConditionalGenerationConfig(RBLNDecoderOnlyModelForCausalLMConfig):
+class DeprecatedLoadVisualRuntimeMixin:
```
rebel-seinpark, line 22: How about move into `optimum-rbln/src/optimum/rbln/utils/deprecation.py`?

rebel-jongho: While this cannot be reused or generalized, `rbln/utils/` refers to general-purpose tools.
This is intended solely for qwen vl.

PR #761, opened 2026-09-17
refactor: use config.get_text_config() to reuse redundant methods

rebel-dkhong:
# Pull Request Description

> 

## Type of Change

- [ ] New Model Support
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

### AS-IS

HuggingFace `transformers` configs come in two shapes
- flat (`LlamaConfig`, `Qwen3Config`)
  - the decoder hyperparameters live at the top level
- composite (`Qwen2_5_VLConfig`, `Qwen3ASRConfig`, `Gemma3Config`)
  - they live only under `text_config`

The shared decoder-only implementation reads them off the flat shape (`self.config.hidden_size`, `model_config.max_position_embeddings`, …), so every composite-config port had to override those methods just to re-read the same values from `config.text_config`.

### TO-BE

HuggingFace `transformers` provides a [`get_text_config()`](https://huggingface.co/docs/transformers/main_classes/configuration#transformers.PreTrainedConfig.get_text_config) method on config classes.

From its docstring:
```
1. On most models, it returns the original config instance itself.
2. On newer (2024+) composite models, it returns the text section of the config, which is nested under a set of valid names.
```

So a flat config returns itself and a composite config returns its text section.

Routing the shared path through it removes the need for overrides such as `_create_embedding_layer`.

## Related Issues

----
# Conventional commit
```
```
----
# Type candidate
  - Model Updates
      - ex) Add LlavaNext 
      - ex) Bugfix Whisper
  - Enhancements
  - Code Refactor
      - ex) Refactor Seq2Seq
  - Documentation
  - Library Dependencies
  - Release
      - ex) Release v1.2.0
  - Other
    - `other`: None of above.
      - ex) ci update
      - ex) pdm update


src/optimum/rbln/transformers/models/qwen3_asr/modeling_qwen3_asr.py
```diff
@@ -302,74 +270,6 @@ def _merge_audio_embeds(
...
-        input_ids: torch.LongTensor | None,
-        inputs_embeds: torch.Tensor | None,
-        cache_position: torch.Tensor | None,
-        attention_mask: torch.LongTensor | None,
-        generate_idx: torch.Tensor | None,
-        padded_cache_lengths: torch.Tensor | None,
-        position_ids: torch.Tensor | None,
-        token_type_ids: torch.Tensor | None,
-        lora_int_ids: torch.Tensor | None,
-    ) -> dict:
```
rebel-jongho, line 316: was it just dead code?

rebel-dkhong: `Qwen3ASRConfig` is a composite config: `hidden_size` and `num_hidden_layers` exist only under `text_config`. 
So the base implementation's `self.config.hidden_size` / `self.config.num_hidden_layers` raised `AttributeError`.

Those two reads sit inside the `if self.rbln_config.output_hidden_states` branch, so the default path went through the base fine and only `output_hidden_states=True` failed. The override existed to read the same values from `self.config.get_text_config()` instead.

Now that the base resolves the text config itself, the override is redundant and has been removed.

